PR #74, opened 2017-11-07
Configuration corrections for running tests.

Xosmond:


paulcsmith:
Looks perfect. Thanks for looking into this

PR #75, opened 2017-11-07
Raises RecordNotFound in find query if no record is found

Xosmond:
I took ActiveRecord's RecordNotFound Error as example for this PR. If you have any suggestions, tell me.

### DIFF
--- a/src/lucky_record/errors.cr
+++ b/src/lucky_record/errors.cr
@@ -1,0 +1,20 @@
+# = Lucky Record Errors
+#
+# Generic Lucky Record exception class.
+class LuckyRecordError < Exception
+end
+
+# Raised when Lucky Record cannot find a record by given id
+class RecordNotFound < LuckyRecordError
+  getter model : (String|Nil)
+  getter primary_key : (String|Nil)
+  getter id : (String|Int32|Nil)
+
+  def initialize(message : (String|Nil) = nil, model : (String|Nil) = nil, primary_key : (String|Nil) = nil, id : (String|Int32|Nil) = nil)
+    @primary_key = primary_key
+    @model = model
+    @id = id
+
+    super(message)
+  end
+end
--- a/src/lucky_record/errors.cr
+++ b/src/lucky_record/errors.cr
@@ -7,7 +7,7 @@ module LuckyRecord
 
   # Raised when Lucky Record cannot find a record by given id
   class RecordNotFoundError < LuckyRecordError
-    def initialize(@model : Symbol, @id : Int32)
+    def initialize(@model : Symbol, @id : String)
       super "Could not find #{model} with id of #{id}"
     end
   end
--- a/src/lucky_record/errors.cr
+++ b/src/lucky_record/errors.cr
@@ -1,20 +1,14 @@
-# = Lucky Record Errors
-#
-# Generic Lucky Record exception class.
-class LuckyRecordError < Exception
-end
+module LuckyRecord
+  # = Lucky Record Errors
+  #
+  # Generic Lucky Record exception class.
+  class LuckyRecordError < Exception
+  end
 
-# Raised when Lucky Record cannot find a record by given id
-class RecordNotFound < LuckyRecordError
-  getter model : (String|Nil)
-  getter primary_key : (String|Nil)
-  getter id : (String|Int32|Nil)
-
-  def initialize(message : (String|Nil) = nil, model : (String|Nil) = nil, primary_key : (String|Nil) = nil, id : (String|Int32|Nil) = nil)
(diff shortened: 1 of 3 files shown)


src/lucky_record/errors.cr
```diff
@@ -0,0 +1,20 @@
+# = Lucky Record Errors
+#
+# Generic Lucky Record exception class.
+class LuckyRecordError < Exception
+end
+
+# Raised when Lucky Record cannot find a record by given id
+class RecordNotFound < LuckyRecordError
```
paulcsmith: Could you nest this in a `LuckyRecord` module? `class LuckyRecord::RecordNotFoundError`

Oh and it should end with `Error`. I think that's the Crystal convention

src/lucky_record/errors.cr
```diff
@@ -0,0 +1,20 @@
...
+  getter model : (String|Nil)
+  getter primary_key : (String|Nil)
+  getter id : (String|Int32|Nil)
```
paulcsmith: I think these should not be nilable. That means you'd need to pass them in when you instantiate the Error. Also I think we can remove the `primary_key` field for now since it is always `id`

src/lucky_record/errors.cr
```diff
@@ -0,0 +1,20 @@
...
+
+  def initialize(message : (String|Nil) = nil, model : (String|Nil) = nil, primary_key : (String|Nil) = nil, id : (String|Int32|Nil) = nil)
```
paulcsmith: In Crystal you can set the instance variable right away like this: `def initialize(@model, @id)`. You also don't need to redeclare the types since you did that in the getter

src/lucky_record/errors.cr
```diff
@@ -0,0 +1,20 @@
...
+    @primary_key = primary_key
+    @model = model
+    @id = id
+
+    super(message)
```
paulcsmith: Might be cool to make a custom message like `Could not find #{model} with id of #{id}`

You can see here that I did that in Habitat by calling super with a string: https://github.com/luckyframework/habitat/blob/master/src/habitat.cr#L4

So this would look like `super "Could not find #{model} with id of #{id}"`

src/lucky_record/queryable.cr
```diff
@@ -29,7 +29,7 @@ module LuckyRecord::Queryable(T)
   end
 
   def find(id)
-    id(id).limit(1).first
+    id(id).limit(1).first? || raise RecordNotFound.new
```
paulcsmith: Maybe you can pass in just the table name and the id: `RecordNotFound.new(id: id, table_name: @@table_name)`. That's probably all the error needs for now

src/lucky_record/errors.cr
Xosmond: It would be better only `LuckyRecord::RecordNotFoundError `or `LuckyRecord::Errors::RecordNotFoundError`?

Im wrapping the classes with the module instead of each line.

src/lucky_record/errors.cr
```diff
@@ -0,0 +1,14 @@
+module LuckyRecord
+  # = Lucky Record Errors
+  #
+  # Generic Lucky Record exception class.
+  class LuckyRecordError < Exception
+  end
+
+  # Raised when Lucky Record cannot find a record by given id
+  class RecordNotFoundError < LuckyRecordError
+    def initialize(@model : Symbol, @id : Int32)
```
paulcsmith: One last thing. Could you make the id also accept a string. Right now it’s possible to use a String because LuckyRecord casts it to an Int32. 

You can add a test for this by duplicating what you have but passing “id” in.

Xosmond: If i pass "id" I get <PQ::PQError: invalid input syntax for integer: "id">, only if I pass "1" RecordNotFoundError will be raised, so should I make the test with "1"?

paulcsmith: Hmm that's a good question. I'm going to need to pull this down and take a look I think. If I don't get back to you in a couple days can you ping me to remind me to take a look?

Xosmond: Sure, no problem, I think we need to try to cast before and raise the correct error before PG does.

paulcsmith: It should cast when you call `id`. I think for now what we can do is accept a string in the Error and call `...Error.new(id: id.to_s)` to make sure it is a String when it gets to the error. In the long term I'm going to make it so that there is more type restriction, but until that's done this should do the trick.

```crystal
... || LuckyRecord::RecordNotFoundError.new(model: @@table_name, id: id.to_s)
```

PR #82, opened 2017-11-10
Add `!` versions of save and update

### LINKED ISSUE #66: Add `!` versions of save and update
Raise if it fails to save/update

```crystal
MyForm.save! params # Returns the record or raises if it fails to validate or save
MyForm.update! params # Returns the record or raises if it fails to validate or save
```

This would add a method to the `LuckyRecord::Form` called `self.save!` and `self.update!`. They would work similar to the non `!` methods, but would not accept a block. They would instead return the record.

edwardloveall:
Fixes #66 

These methods will work like the normal `save` and `update`, but only take parameters and no block. If it saves successfully, it will return the record, _not_ the form. If it fails, it will raise a LuckyRecord::Exceptions::InvalidForm error.

I tried to use the existing logic where it made sense, since the goal is for the method to work the same but have some different behavior after the record is saved (or not).

I took the hierarchy pattern for the error from LuckyWeb's exception module.

### DIFF
--- a/spec/form_spec.cr
+++ b/spec/form_spec.cr
@@ -243,7 +243,10 @@ describe "LuckyRecord::Form" do
       it "raises an exception" do
         params = {"name" => "", "age" => "30"}
 
-        expect_raises(LuckyRecord::InvalidFormError) do
+        expect_raises(
+          LuckyRecord::InvalidFormError(UserForm),
+          /Invalid UserForm. Could not save/
+        ) do
           UserForm.save!(params)
         end
       end
@@ -296,7 +299,10 @@ describe "LuckyRecord::Form" do
         user = UserQuery.new.first
         params = {"name" => ""}
 
-        expect_raises(LuckyRecord::InvalidFormError) do
+        expect_raises(
+          LuckyRecord::InvalidFormError(UserForm),
+          /Invalid UserForm. Could not save/
+        ) do
           UserForm.update! user, with: params
         end
       end
--- a/spec/form_spec.cr
+++ b/spec/form_spec.cr
@@ -227,6 +227,29 @@ describe "LuckyRecord::Form" do
     end
   end
 
+  describe ".save!" do
+    context "on success" do
+      it "saves and returns the record" do
+        params = {"joined_at" => now_as_string, "name" => "New Name", "age" => "30"}
+
+        record = UserForm.save!(params)
+
+        record.is_a?(User).should be_true
+        record.name.should eq "New Name"
+      end
+    end
+
+    context "on failure" do
+      it "raises an exception" do
+        params = {"name" => "", "age" => "30"}
+
+        expect_raises(LuckyRecord::InvalidFormError) do
+          UserForm.save!(params)
+        end
+      end
+    end
+  end
+
   describe ".update" do
     context "on success" do
       it "yields the form and the updated record" do
--- a/spec/form_spec.cr
+++ b/spec/form_spec.cr
@@ -275,4 +275,31 @@ describe "LuckyRecord::Form" do
       end
     end
   end
+
+  describe ".update!" do
+    context "on success" do
+      it "updates and returns the record" do
+        create_user(name: "Old Name")
+        user = UserQuery.new.first
+        params = {"name" => "New Name"}
+
+        record = UserForm.update! user, with: params
+
+        record.is_a?(User).should be_true
+        record.name.should eq "New Name"
+      end
+    end
+
+    context "on failure" do
+      it "raises an exception" do
+        create_user(name: "Old Name")
+        user = UserQuery.new.first
+        params = {"name" => ""}
+
+        expect_raises(LuckyRecord::InvalidFormError) do
+          UserForm.update! user, with: params
+        end
+      end
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


src/lucky_record/exceptions.cr
```diff
@@ -0,0 +1,9 @@
+module LuckyRecord
+  module Exceptions
+    class Base < Exception
+    end
+
+    class InvalidForm < Base
```
edwardloveall: I kinda guessed on an appropriate class name here.

paulcsmith: I think InvalidForm is a good name. It can go here with the other errors: https://github.com/luckyframework/record/blob/master/src/lucky_record/errors.cr

Also need `Error` at the end. Might be worth passing in the form name and form object as well? That way it can be used in the output or for automatically handling form errors in JSON APIs

src/lucky_record/form.cr
```diff
@@ -167,6 +167,18 @@ abstract class LuckyRecord::Form(T)
     end
   end
 
+  def save! : T
+    if save
+      record.as(T)
```
paulcsmith: I think the idea here was to tell the compiler that it is not nil right? If so you can call `not_nil!` on it

src/lucky_record/needy_initializer.cr
```diff
@@ -53,6 +53,20 @@ module LuckyRecord::NeedyInitializer
...
+        {% if NEEDS.size > 0 %}
+          **needs
+        {% end %}
+      )
+      form = new(
+        params,
+        {% if NEEDS.size > 0 %}
+          **needs
+        {% end %}
+      ).save!
```
paulcsmith: At some point I think it would be good to extract a `run` macro https://crystal-lang.org/api/0.23.1/Crystal/Macros.html#run%28filename%2C%2Aargs%29%3AMacroId-instance-method so that we don't need to repeat the arguments list every time for each version of save. But I think this is fine for now. Here's another example of a run macro in case you're curious: https://github.com/luckyframework/web/blob/master/src/run_macros/infer_route.cr called from https://github.com/luckyframework/web/blob/master/src/lucky_web/routeable.cr#L47

edwardloveall: Yeah this started to feel not DRY. Also, I need to read more about macros.

src/lucky_record/form.cr
```diff
@@ -167,6 +167,18 @@ abstract class LuckyRecord::Form(T)
     end
   end
 
+  def save! : T
+    if save
+      record.not_nil!
+    else
+      raise LuckyRecord::InvalidFormError.new("#{record} has invalid parameters")
```
edwardloveall: I'm not quite sure how you are thinking I pass and store the form name/object. Did you mean something like this:

```crystal
class InvalidFormError(T) < Exception
  getter form_name : String
  getter form_object : T

  def initialize(@form_name : String, @form_object : T)
    super(message: "Error in #{form_name}")
  end
end
```

My Crystal's not great yet. Hopefully you understand what I mean.

paulcsmith: It would look something like this: https://github.com/luckyframework/record/commit/576cd3a440b6bb5095594e607198f5c0d717433c#diff-0c6732b55ddd3e3a8985f88608914664R32

And yeah I'd add getting for the form_name and form_object. You don't need to specify the type in `getter` if it is specific in the initializer. LMK if you have any other questions though. Happy to help :)

edwardloveall:
I tried swapping out various permutations of `T` for `LuckyWeb::Form` as the type and couldn't get it work. I kept getting a `use a more specific type` error. If you can figure it out I'll happily change it 😄

edwardloveall:
Also, passing `record` doesn't seem to work very well because it's `nil` if the form hasn't been saved yet. Not quite sure what would be best in that case.

src/lucky_record/form.cr
```diff
@@ -167,6 +167,18 @@ abstract class LuckyRecord::Form(T)
     end
   end
 
+  def save! : T
+    if save
+      record.not_nil!
+    else
+      raise LuckyRecord::InvalidFormError.new(form_name: typeof(self).to_s, form_object: record)
```
paulcsmith: Instead of passing record for the form object, I think it should be `self` so that it returns the form object. And I think there is a `form_name` method you can use so you don't need to do `typeof`

paulcsmith: Actually you're right. typeof(self).to_s is probably better because `form_name` will return the underscored version. So that can probably be kept as-is 👍

spec/form_spec.cr
```diff
@@ -227,6 +227,29 @@ describe "LuckyRecord::Form" do
...
+        record.is_a?(User).should be_true
+        record.name.should eq "New Name"
+      end
+    end
+
+    context "on failure" do
+      it "raises an exception" do
+        params = {"name" => "", "age" => "30"}
+
+        expect_raises(LuckyRecord::InvalidFormError) do
```
paulcsmith: You can pass a string as the second arg to `expect_raises` and it will check that it is included in the error message. So maybe it would be good to add "Invalid UserForm"

edwardloveall:
Done! 😄 I had to use a regex for the exception message matching because the message prints the full instance variable with address:

```
LuckyRecord::InvalidFormError(UserForm): Invalid UserForm. Could not save #<UserForm:0x107c753c0>
```

paulcsmith:
Looks great. This will really help open up some new use cases for forms 👍

PR #85, opened 2017-11-12
Custom messages for validations

actsasflinn:
Adds the ability to customize validation messages as a String, Proc or Object#call


spec/validations_spec.cr
```diff
@@ -21,6 +34,17 @@ private class TestValidationUser
     self
   end
 
+  def run_validations_with_message
+    validate_required city, state, message: "ugh"
+    validate_inclusions_of state, in: ["CA, NY"], message: "that one's not allowed"
+    validate_confirmation_of name, message: "name confirmation much match"
```
paulcsmith: much -> must

spec/validations_spec.cr
```diff
@@ -21,6 +34,17 @@ private class TestValidationUser
...
+  end
+
+  def run_validations_with_message_callables
+    validate_required city, state, message: ->(x : String, y : String){ "#{x} required message from Proc" }
```
paulcsmith: I like the idea of allowing a Proc or class, but I think making the names order dependent like this could be a bit confusing and/or lead to bugs. I'm thinking that if you want to use a proc for `validate_required` then you must do it with just one field at a time. 

```crystal
validate_required city, message: ->(city : String) { "#{city} required message from Proc"}
```

I think allowing a string message is fine with an array of fields though since that makes sense and wouldn't break if you change the field order.

The restriction would look like this I think:

```crystal
# Change the existing method signature
def validate_required(**fields, message : String = "is required")

# Add a new one for use with just one field 
def validate_required(field, message : Object)
```

src/lucky_record/validations.cr
```diff
@@ -1,27 +1,27 @@
...
-  macro validate_confirmation_of(field_name)
+  macro validate_confirmation_of(field_name, message = "must match")
     if {{ field_name.id }}.value != {{ field_name.id }}_confirmation.value
-      {{ field_name }}.add_error "must match"
+      {{ field_name }}.add_error {{ message }}
     end
   end
 
-  private def validate_inclusions_of(field, in allowed_values)
+  private def validate_inclusions_of(field, in allowed_values, message = "is invalid")
```
paulcsmith: I just realized there is a typo here. Oops :)

src/lucky_record/field.cr
```diff
@@ -12,7 +12,8 @@ class LuckyRecord::Field(T)
     @param || value.to_s
   end
 
-  def add_error(message : String)
+  def add_error(message : Object = "is invalid")
+    message = message.call(@name.to_s, @value.to_s) if message.responds_to?(:call)
```
paulcsmith: `responds_to` makes it less clear what this object needs and could potentially lead to errors. There is a not super well documented feature that we can use to make it more clear what this accepts without relying on `responds_to?`

When a class includes a module, you can restrict it by that type. So I think we can do something like this

```crystal
module LuckyRecord::CallableErrorMessage
  abstract def call(name, value)
end

class CallableMessage
  include LuckyRecord::CallableErrorMessage
end

# Then `add_error` changes to:
def add_error(message : String)
# And adds an additional one
def add_error(message : (Proc | LuckyRecord::CallableErrorMessage))
```

src/lucky_record/validations.cr
paulcsmith: https://github.com/luckyframework/record/issues/95

spec/validations_spec.cr
actsasflinn: It might look more confusing since I called them x and y in the spec. The Proc arguments aren't city and state, they are the name and value relative to the field being validated. Let me know if I've interpreted your comment right. I can make the spec more clear on what it's doing by renaming the Proc args.

src/lucky_record/field.cr
actsasflinn: I think I like that. It makes it more clear the requirement for name and value.I have to admit I tried hacking together ways to provide some flexibility providing a proc or callable with 0 args, 1 arg and 2 args... but it was gnarly. I'm not sure if there's a way to do this properly in

spec/validations_spec.cr
```diff
@@ -21,6 +35,17 @@ private class TestValidationUser
...
   end
 
+  def run_validations_with_message
+    validate_required city, state, message: "ugh"
+    validate_inclusion_of state, in: ["CA, NY"], message: "that one's not allowed"
+    validate_confirmation_of name, with: name_confirmation, message: "name confirmation must match"    
+  end
+
+  def run_validations_with_message_callables
+    validate_required city, state, message: ->(field_name : String, field_value : String){ "#{field_name} required message from Proc" }
```
paulcsmith: This makes more sense now 👍

spec/validations_spec.cr
```diff
@@ -21,6 +35,17 @@ private class TestValidationUser
...
+    validate_inclusions_of state, in: ["CA, NY"], message: CallableMessage.new(@name)
```
paulcsmith: I think if you rebase this will be `validate_inclusion_of` without the "s" in inclusions

src/lucky_record/validations/callable_error_message.cr
```diff
@@ -0,0 +1,3 @@
+module LuckyRecord::CallableErrorMessage
+  abstract def call(name, value)
+end
```
paulcsmith: Missing newline here

Issue #90, opened 2017-11-12
Only select fields that the model defines


paulcsmith:
I will probably tackle this at some point, but if someone wants to give it a shot please comment here

The idea would be to change:
https://github.com/luckyframework/record/blob/5852d4b057b8f03544a6d2f231678834542a19c3/src/lucky_record/query_builder.cr#L78-L80

To something like:

```
SELECT user.name, user.age, user.email # etc. where the fields come from the fields defined on the model
```

hanneskaeufler:
Had my fun with this in #106, feel free to pick it apart or scrap it if this is not what you had in mind!

paulcsmith:
Closed by #106. Thanks @hanneskaeufler!

PR #105, opened 2017-11-18
Add aggregate 'count' to query builder/queryable

hanneskaeufler:
This should add support for queries like `UserQuery.new.company.lower.is("thoughtbot").count`. Forgive me if anything is absolutely misplaced, just finding my way around in here :) I guess this also has some relation to #90.

Edit:

Should the `LuckyRecord::QueryBuilder` somehow be "closed" after calling `.count`?
I guess things like
```crystal
    query = LuckyRecord::QueryBuilder
      .new(table: :users)
      .count
      .order_by(:email, :desc)
```
wouldn't make too much sense?

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -80,6 +80,17 @@ describe LuckyRecord::Query do
       query.statement.should eq "SELECT * FROM users ORDER BY name ASC"
     end
   end
+
+  describe "#count" do
+    it "returns the number of database rows" do
+      count = UserQuery.new.count
+      count.should eq 0
+
+      insert_a_user
+      count = UserQuery.new.count
+      count.should eq 1
+    end
+  end
 end
 
 private def insert_a_user
--- a/src/lucky_record/query_builder.cr
+++ b/src/lucky_record/query_builder.cr
@@ -6,7 +6,7 @@ class LuckyRecord::QueryBuilder
     asc:  [] of Symbol | String,
     desc: [] of Symbol | String,
   }
-  @selections = "*"
+  @selections : String = "*"
   @prepared_statement_placeholder = 0
 
   VALID_DIRECTIONS = [:asc, :desc]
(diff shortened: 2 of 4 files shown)


spec/query_spec.cr
```diff
@@ -80,6 +80,17 @@ describe LuckyRecord::Query do
...
   end
+
+  describe "#count" do
+    it "returns the number of database rows" do
+      count = UserQuery.new.count
+      count.should eq 0
+
+      insert_a_user
+      count = UserQuery.new.count
+      count.should eq 1
```
paulcsmith: Nice test. This looks perfect

src/lucky_record/query_builder.cr
```diff
@@ -6,6 +6,7 @@ class LuckyRecord::QueryBuilder
     asc:  [] of Symbol | String,
     desc: [] of Symbol | String,
   }
+  @selections = "*"
```
paulcsmith: Could you add a specific type to this `@selections : String = "*"`?

src/lucky_record/queryable.cr
```diff
@@ -37,6 +37,11 @@ module LuckyRecord::Queryable(T)
     exec_query.first
   end
 
+  def count
+    query.count
+    exec_scalar
```
paulcsmith: Can you do `as(Int64)` on this so it's clear what it returns?

src/lucky_record/queryable.cr
```diff
@@ -37,6 +37,11 @@ module LuckyRecord::Queryable(T)
     exec_query.first
   end
 
+  def count
```
paulcsmith: And `def count : Int64` so we can be sure it's an Int64

paulcsmith:
Regarding your last comment, I think it should have some idea of being "closed" but I'm not sure exactly what that should look like. For now this is ok, especially since users will rarely use QueryBuilder directly

hanneskaeufler:
Comments addressed, thanks for the tips!

paulcsmith:
Looks awesome. Thanks :)

PR #106, opened 2017-11-18
Hk select specific fields

hanneskaeufler:
This tries to adress #90. Most of the commit(s) is noise regarding to the fact that we have no more `*` in the query by default, but the models specific fields.

How did I go on about this?
There is a macro-defined class method `fields` on `LuckyRecord::Model` which returns all the fields defined by the `field` macro. So when a query is built for a model, the `selection` method on the `QueryBuilder` is called with all of those field.

Defining a class method `fields` means a user can never add a class method `fields` himself I guess? Also, using `.selection([:foo, :bar])` is by no means safe, could easily blow up on execution, right?


spec/criteria_spec.cr
```diff
@@ -9,41 +9,45 @@ end
...
+      age.lte("30").to_sql.should eq ["SELECT #{query_me_columns} FROM users WHERE age <= $1", "30"]
     end
   end
 end
 
 private def age
   QueryMe::BaseQuery.new.age
 end
+
+private def query_me_columns
```
paulcsmith: I think since this is repeated a bit, I think it might be good to add a `COLUMNS` constant to the QueryMe class so you can do `QueryMe::COLUMNS` instead of defining this helper in multiple specs. What do you think?

spec/model_spec.cr
```diff
@@ -84,4 +84,14 @@ describe LuckyRecord::Model do
 
     User::BaseQuery.all.size.should eq 0
   end
+
+  describe ".fields" do
```
paulcsmith: I think we should name this `column_names` so it's more clear that they aren't `LuckyRecord::Field`s

spec/query_builder_spec.cr
```diff
@@ -55,6 +55,24 @@ describe "LuckyRecord::QueryBuilder" do
     query.statement.should eq "SELECT COUNT(*) FROM users"
     query.args.should eq [] of String
   end
+
+  describe "#select" do
+    it "can specify a column to be selected" do
```
paulcsmith: I think this spec can be removed since the one below will cover both cases automatically

spec/query_spec.cr
```diff
@@ -103,3 +103,7 @@ private def insert_a_user
       Time.now
   end
 end
+
+private def user_columns
```
paulcsmith: Same comment here about extracting a `COLUMNS` to the query class

src/lucky_record/model.cr
```diff
@@ -86,4 +87,14 @@ class LuckyRecord::Model
     {% end %}
     {% FIELDS << {name: type_declaration.var, type: data_type, nilable: nilable.id, autogenerated: autogenerated} %}
   end
+
+  macro setup_fields_method
+    def self.fields
```
paulcsmith: Can you add a return type of `def self.fields : Array(Symbol)`? I know a lot of existing code doesn't do that, but I'd like to do it more so that the generated API docs are better

src/lucky_record/query_builder.cr
```diff
@@ -75,6 +75,13 @@ class LuckyRecord::QueryBuilder
     self
   end
 
+  def select(selection : Array(Symbol))
```
paulcsmith: I like this type restriction 👍 That way we can later add a `select(selection : String)` that can be used to select just about anything

spec/criteria_spec.cr
hanneskaeufler: `QueryMe` is defined multiple times with slight differences, so cannot quite remove it. But I agree, a `COLUMNS` puts it at a better place. Done ✅

spec/model_spec.cr
hanneskaeufler: Ah mhm wasn't quite aware of `Field` being already used somewhere, thanks. ✅

spec/query_builder_spec.cr
hanneskaeufler: Yeah, it just moved me along tdd'ing this. Never afraid to remove specs 😈 ✅

src/lucky_record/model.cr
hanneskaeufler: Yepp, am new to crystal so have no real feeling for when to add types and when to infer them yet. ✅

spec/bool_criteria_spec.cr
```diff
@@ -1,6 +1,8 @@
 require "./spec_helper"
 
 private class QueryMe < LuckyRecord::Model
+  COLUMNS = "users.id, users.created_at, users.updated_at, users.admin"
```
paulcsmith: Nice!

hanneskaeufler:
Thanks for the review, I like the improvements!

paulcsmith:
Circle wasn't set up to run forks. I changed that setting and will open/close this so it runs in CI

PR #110, opened 2017-11-19
Add support for NOT operator, closes #4

### LINKED ISSUE #4: Add support for NOT operator

hanneskaeufler:
~~This allows us to say `UserQuery.new.where.not(:name, "Paul")`, meaning
a negation of the condition that is passed into `not`.~~

Edit:
As per your gist

```crystal
UserQuery.new.age.not("30")
UserQuery.new.name.not.ilike("peter")
```

is now possible.

~~#inspired-by-active-record~~

### DIFF
--- a/spec/criteria_spec.cr
+++ b/spec/criteria_spec.cr
@@ -46,13 +46,13 @@ describe LuckyRecord::Criteria do
   end
 
   describe "not" do
-    describe "with a value" do
+    describe "without chained criteria" do
       it "negates to not equal" do
         age.not("30").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age != $1", "30"]
       end
     end
 
-    describe "without a value" do
+    describe "with chained criteria" do
       it "negates the following criteria" do
         age.not.ilike("3").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age NOT ILIKE $1", "3"]
       end
--- a/spec/criteria_spec.cr
+++ b/spec/criteria_spec.cr
@@ -44,6 +44,24 @@ describe LuckyRecord::Criteria do
       age.lte("30").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age <= $1", "30"]
     end
   end
+
+  describe "not" do
+    describe "with a value" do
+      it "negates to not equal" do
+        age.not("30").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age != $1", "30"]
+      end
+    end
+
+    describe "without a value" do
+      it "negates the following criteria" do
+        age.not.ilike("3").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age NOT ILIKE $1", "3"]
+      end
+    end
+
+    it "it resets after having negated once" do
+      age.not.like("3").age.lt("20").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age NOT LIKE $1 AND age < $2", "3", "20"]
+    end
+  end
 end
 
 private def age
--- a/spec/criteria_spec.cr
+++ b/spec/criteria_spec.cr
@@ -56,10 +56,10 @@ describe LuckyRecord::Criteria do
       it "negates the following criteria" do
         age.not.gt("3").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age <= $1", "3"]
       end
-    end
 
-    it "it resets after having negated once" do
-      age.not.gt("3").age.is("20").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age <= $1 AND age = $2", "3", "20"]
+      it "resets after having negated once" do
+        age.not.gt("3").age.is("20").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age <= $1 AND age = $2", "3", "20"]
+      end
     end
   end
 end
--- a/spec/criteria_spec.cr
+++ b/spec/criteria_spec.cr
@@ -59,7 +59,7 @@ describe LuckyRecord::Criteria do
     end
 
     it "it resets after having negated once" do
-      age.not.like("3").age.lt("20").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age NOT LIKE $1 AND age < $2", "3", "20"]
+      age.not.like("3").age.is("20").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age NOT LIKE $1 AND age = $2", "3", "20"]
     end
   end
 end
--- a/spec/criteria_spec.cr
+++ b/spec/criteria_spec.cr
@@ -54,12 +54,12 @@ describe LuckyRecord::Criteria do
 
     describe "with chained criteria" do
       it "negates the following criteria" do
-        age.not.ilike("3").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age NOT ILIKE $1", "3"]
+        age.not.gt("3").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age <= $1", "3"]
       end
     end
 
     it "it resets after having negated once" do
-      age.not.like("3").age.is("20").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age NOT LIKE $1 AND age = $2", "3", "20"]
+      age.not.gt("3").age.is("20").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age <= $1 AND age = $2", "3", "20"]
     end
   end
 end
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -93,11 +93,21 @@ describe LuckyRecord::Query do
   end
 
   describe "#not" do
-    it "negates the given where condition" do
+    it "negates the given where condition as 'equal'" do
       insert_a_user
-      results = UserQuery.new.where.not(:name, "not the name").results
 
+      results = UserQuery.new.name.not("not the name").results
       results.should eq UserQuery.new.results
+
+      results = UserQuery.new.name.not("Paul").results
+      results.should eq [] of User
+    end
+
+    it "negates any previous condition" do
+      insert_a_user
+
+      results = UserQuery.new.name.not.is("Paul").results
+      results.should eq [] of User
     end
   end
 end
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -91,6 +91,16 @@ describe LuckyRecord::Query do
       count.should eq 1
     end
   end
+
+  describe "#not" do
+    it "negates the given where condition" do
+      insert_a_user
+      query = UserQuery.new.where.not(:name, "not the name").query
+
+      query.statement.should eq "SELECT #{User::COLUMNS} FROM users WHERE name != $1"
+      query.args.should eq ["not the name"]
+    end
+  end
 end
 
 private def insert_a_user
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -92,7 +92,7 @@ describe LuckyRecord::Query do
     end
   end
 
-  describe "#not" do
+  describe "#not with an argument" do
     it "negates the given where condition as 'equal'" do
       insert_a_user(name: "Paul")
 
@@ -107,12 +107,22 @@ describe LuckyRecord::Query do
       results = UserQuery.new.name.lower.not("alex").results
       results.map(&.name).should eq ["Paul", "Sarah"]
     end
+  end
 
+  describe "#not with no arguments" do
     it "negates any previous condition" do
       insert_a_user
 
       results = UserQuery.new.name.not.is("Paul").results
       results.should eq [] of User
+    end
+
+    it "can be used with operators" do
+      insert_a_user(name: "Joyce", age: 33)
+      insert_a_user(name: "Jil", age: 34)
+
+      results = UserQuery.new.age.not.gt(33).results
+      results.map(&.name).should eq ["Joyce"]
     end
   end
 end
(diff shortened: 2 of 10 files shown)


spec/where_spec.cr
```diff
@@ -0,0 +1,93 @@
+require "./spec_helper"
+
+describe LuckyRecord::Where do
+  describe "Equal" do
+    describe "#negated" do
+      it "returns a NotEqual" do
+        LuckyRecord::Where::Equal.new("column", "value").negated
```
hanneskaeufler: All of those `LuckyRecord::Where` tests feel super boilerplate-y and I was unable to verify the name/column instance vars. Thats why it's a `.should be_a`. Do you have any ideas to improve that?

hanneskaeufler: Also, at the moment we could get away with not having any of that, since `Queryable` only uses `Equal` ... hmmm

src/lucky_record/query_builder.cr
```diff
@@ -103,6 +103,15 @@ class LuckyRecord::QueryBuilder
     self
   end
 
+  def where
+    self
+  end
+
+  def not(where_clause : LuckyRecord::Where::SqlClause)
```
hanneskaeufler: This isn't needed as well strictly speaking. Since `Queryable` only uses `Equal` we could get by without having `not` and `where` on the `QueryBuilder`. Am I missing where all the other `SqlClause`es are being used?

hanneskaeufler:
No worries at all. I will try to massage this a bit in order to get closer to a likeable solution!

hanneskaeufler:
Took another look at this and ended up with something closer to the gist you posted. I don't particularly like my implementation, but ah well it passes the current tests 🤷‍♂️

spec/criteria_spec.cr
```diff
@@ -44,6 +44,24 @@ describe LuckyRecord::Criteria do
...
+      end
+    end
+
+    describe "without a value" do
+      it "negates the following criteria" do
+        age.not.ilike("3").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age NOT ILIKE $1", "3"]
+      end
+    end
+
+    it "it resets after having negated once" do
```
paulcsmith: This is a really nice spec. I think it will be great to have in order to be sure we don't have any regressions in the future 👍

spec/criteria_spec.cr
```diff
@@ -44,6 +44,24 @@ describe LuckyRecord::Criteria do
       age.lte("30").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age <= $1", "30"]
     end
   end
+
+  describe "not" do
+    describe "with a value" do
```
paulcsmith: Maybe "without chained criteria"?

spec/criteria_spec.cr
```diff
@@ -44,6 +44,24 @@ describe LuckyRecord::Criteria do
...
+      it "negates to not equal" do
+        age.not("30").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age != $1", "30"]
+      end
+    end
+
+    describe "without a value" do
```
paulcsmith: "with chained criteria"

spec/query_spec.cr
```diff
@@ -91,6 +91,25 @@ describe LuckyRecord::Query do
       count.should eq 1
     end
   end
+
+  describe "#not" do
+    it "negates the given where condition as 'equal'" do
```
paulcsmith: I think the not works with more than just equal and not equal. So I think this description could be `"negates the given where condition"`

spec/query_spec.cr
```diff
@@ -91,6 +91,25 @@ describe LuckyRecord::Query do
...
+      insert_a_user
+
+      results = UserQuery.new.name.not("not the name").results
+      results.should eq UserQuery.new.results
+
+      results = UserQuery.new.name.not("Paul").results
+      results.should eq [] of User
```
paulcsmith: I'd love to see one extra spec with extra changing, and maybe some updates to the `insert_a_user` method to make it more clear why these specs pass. What do you think about this?

```crystal
    it "negates the given where condition" do
      insert_a_user(name: "Paul")

      results = UserQuery.new.name.not("not the name").results
      results.map(&.name).should eq ["Paul"]

      results = UserQuery.new.name.not("Paul").results
      results.should eq [] of User

      insert_a_user(name: "Jill")
      insert_a_user(name: "Steve")
      results = UserQuery.new.name.lower.not("steve").results
      results.map(&.name).should eq ["Paul", "Jill"]
    end

private def insert_a_user(age = 34, name = "Paul")
  LuckyRecord::Repo.run do |db|
    db.exec "INSERT INTO users(name, created_at, updated_at, age, joined_at) VALUES ($1, $2, $3, $4, $5)",
      name,
      Time.now,
      Time.now,
      age,
      Time.now
  end
end
```

spec/where_spec.cr
paulcsmith: I think Queryable will work with all of them. I pulled it down and tried with gt and lt and those all worked. Which I think is cool that it automatically works like that. I agree these tests do seem a bit boilerplate. There should be a `column` and `value` getter defined on the clause so I think you could do something like this:

```crystal
clause = LuckyRecord::Where::NotEqual.new("column", "value").negated
clause.column.should eq "column"
clause.value.should eq "value"

# And since they are all pretty similar, I think there could be just one "it" block and a helper method could be defined

it "negates the clause" do
  should_negate LuckyRecord::Where::NotEqual, with: LuckyRecord::Where::Equal
  should_negate LuckyRecord::Where::Like, with: LuckyRecord::Where::NotLike
end

private def should_negate(original_where, with expected_negated_where)
  clause = original_where.new("column", "value").negated
  clause.column.should eq "column"
  clause.value.should eq "value"
  clause.should be_a(expected_negated_where)
end
```

src/lucky_record/criteria.cr
```diff
@@ -1,7 +1,9 @@
 class LuckyRecord::Criteria(T, V)
   property :rows, :column
+  @negate_next_criteria : Bool
```
paulcsmith: You could add ` = false` to the end of this and I think you wouldn't need the instance var on line 6

src/lucky_record/criteria.cr
```diff
@@ -15,32 +17,58 @@ class LuckyRecord::Criteria(T, V)
...
+    add_where_sql_clause(LuckyRecord::Where::LessThan.new(column, V::Lucky.to_db!(value)))
   end
 
-  def lte(value)
-    rows.query.where(LuckyRecord::Where::LessThanOrEqualTo.new(column, V::Lucky.to_db!(value)))
+  def lte(value) : T
+    add_where_sql_clause(LuckyRecord::Where::LessThanOrEqualTo.new(column, V::Lucky.to_db!(value)))
+  end
+
+  def like(value) : T
```
paulcsmith: I think `like` and `ilike` should only be added to the String criteria. Adding this to the base Criteria means that it will be available for things like strings and time, but that wouldn't actually work for those types. Does that make sense?

src/lucky_record/where.cr
```diff
@@ -16,47 +16,99 @@ module LuckyRecord::Where
...
+      LessThanOrEqualTo.new(@column, @value)
+    end
   end
 
   class GreaterThanOrEqualTo < SqlClause
     def operator
       ">="
     end
+
+    def negated : SqlClause
```
paulcsmith: I love how you handled these. The `negated` idea is great. The only suggestion I have is to change the restricted type to `LessThan` or whatever the parent type is

paulcsmith:
Here is an example of the string criteria https://github.com/luckyframework/record/blob/master/src/lucky_record/charms/string_extensions.cr

src/lucky_record/criteria.cr
hanneskaeufler: I thought so as well, but we get a `this 'initialize' doesn't initialize instance variable '@negate_next_criteria'`

spec/query_spec.cr
hanneskaeufler: Hmm if you are using `name.not("foo")` it is hardcoded to an `EqualCriteria`. Doing an e.g. `ilike` is only possible using the chained version `name.not.ilike("foo")`. Please correct me if I'm wrong!

spec/where_spec.cr
```diff
@@ -0,0 +1,23 @@
...
+    should_negate(LuckyRecord::Where::Equal, LuckyRecord::Where::NotEqual)
+    should_negate(LuckyRecord::Where::NotEqual, LuckyRecord::Where::Equal)
+    should_negate(LuckyRecord::Where::GreaterThan, LuckyRecord::Where::LessThanOrEqualTo)
+    should_negate(LuckyRecord::Where::GreaterThanOrEqualTo, LuckyRecord::Where::LessThan)
+    should_negate(LuckyRecord::Where::LessThan, LuckyRecord::Where::GreaterThanOrEqualTo)
+    should_negate(LuckyRecord::Where::LessThanOrEqualTo, LuckyRecord::Where::GreaterThan)
+    should_negate(LuckyRecord::Where::Like, LuckyRecord::Where::NotLike)
+    should_negate(LuckyRecord::Where::NotLike, LuckyRecord::Where::Like)
+    should_negate(LuckyRecord::Where::Ilike, LuckyRecord::Where::NotIlike)
+    should_negate(LuckyRecord::Where::NotIlike, LuckyRecord::Where::Ilike)
```
hanneskaeufler: Woah, I could not make your suggestion work for the life of me. As soon as I put the `should be_a()` into a `private def` it fails to compile with 

```
macro didn't expand to a valid program, it expanded to:

================================================================================
--------------------------------------------------------------------------------
   1.       Spec::BeAExpectation(expected_negated_where).new
   2.
--------------------------------------------------------------------------------
Syntax error in expanded macro: be_a:1: expecting token 'CONST', not 'expected_negated_where'

      Spec::BeAExpectation(expected_negated_where).new
```
I sure might be doing something wrong, or am I stumbling on a known issue (https://github.com/crystal-lang/crystal/issues/236)? Solved it with a macro as well, not really sure on the pitfalls/benefits at this point. At least the specs are less boilerplate haha.

hanneskaeufler:
Feel free to have another look, I am reasonably happy with this now! Thanks a bunch for all the input!!! 💟

spec/query_spec.cr
paulcsmith: Oh I see. I didn't realize this was just for `not` without changing anything after. I think changing the describe would make it more clear. Maybe `describe "#not with an argument"` Thoughts?

spec/query_spec.cr
```diff
@@ -91,15 +91,39 @@ describe LuckyRecord::Query do
...
+    end
+
+    it "negates any previous condition" do
+      insert_a_user
+
+      results = UserQuery.new.name.not.is("Paul").results
+      results.should eq [] of User
+    end
+  end
 end
```
paulcsmith: I'd love to see just one more spec like `it "can use not with operators"` that tested `gt` for example. Just testing one is enough to make sure it works. Something along these lines:

```crystal
it "can use not with operators" do
 insert_a_user(age: 28, name: "John")
 insert_a_user(age: 300, name: "Yoda")

  UserQuery.new.age.not.gt(28).results.map(&.name).should eq ["John"]
end
```

spec/where_spec.cr
paulcsmith: Ah this makes sense. I didn't realize `be_a` is a macro so this works 👍

src/lucky_record/charms/string_extensions.cr
```diff
@@ -15,14 +15,12 @@ class String
...
-        rows
+      def like(value : String) : T
+        add_where_sql_clause(LuckyRecord::Where::Like.new(column, value))
       end
 
-      def ilike(value : String)
-        rows.query.where(LuckyRecord::Where::Ilike.new(column, value))
-        rows
+      def ilike(value : String) : T
+        add_where_sql_clause(LuckyRecord::Where::Ilike.new(column, value))
```
paulcsmith: 👍

src/lucky_record/criteria.cr
paulcsmith: 👍

src/lucky_record/criteria.cr
```diff
@@ -15,32 +17,50 @@ class LuckyRecord::Criteria(T, V)
...
-    rows.query.where(LuckyRecord::Where::LessThanOrEqualTo.new(column, V::Lucky.to_db!(value)))
+  def lt(value) : T
+    add_where_sql_clause(LuckyRecord::Where::LessThan.new(column, V::Lucky.to_db!(value)))
+  end
+
+  def lte(value) : T
+    add_where_sql_clause(LuckyRecord::Where::LessThanOrEqualTo.new(column, V::Lucky.to_db!(value)))
+  end
+
+  private def add_where_sql_clause(sql_clause)
```
paulcsmith: Tiny nitpick but maybe this could be `add_clause`? I think that's still clear and a bit more concise

src/lucky_record/charms/string_extensions.cr
```diff
@@ -15,14 +15,12 @@ class String
       @upper = false
       @lower = false
 
-      def like(value : String)
-        rows.query.where(LuckyRecord::Where::Like.new(column, value))
-        rows
+      def like(value : String) : T
```
paulcsmith: A test for the negation of these might be good. Maybe in `spec/string_critiera_spec.cr

hanneskaeufler: These tests already exist. See `spec/string_criteria_spec.cr` exactly 🍾

paulcsmith: Oops! I should look more closely! :D

spec/query_spec.cr
hanneskaeufler: Done ✅

src/lucky_record/criteria.cr
hanneskaeufler: I can live with that, done ✅

hanneskaeufler:
Worked through your comments once more ♻️ Thanks!

spec/criteria_spec.cr
```diff
@@ -44,6 +44,24 @@ describe LuckyRecord::Criteria do
       age.lte("30").to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE age <= $1", "30"]
     end
   end
+
+  describe "not" do
+    describe "without chained criteria" do
```
jdbann: As you're defining conditions for `not` to be run under (with or without chained criteria) I'd use `context` blocks here. Makes it a bit easier to scan back through previous tests, and they are used in other parts of the framework in this way https://github.com/luckyframework/record/blob/master/spec/form_spec.cr

PR #115, opened 2017-11-21
Virtual forms

### LINKED ISSUE #55: Add non-model/virtual forms
```crystal
class SignInForm < LuckyRecord::VirtualForm
  allow_virtual email : String
  allow_virtual password : String

  def submit
    if valid_credentials?
      yield self, user_from_email
    else
      yield self, nil
    end
  end

  private def valid_credentials?
    if user_from_email && !passwords_match?
      password.add_error "is incorrect"
    elsif user_from_email.nil?
      email.add_error "was not found"
    end
  end

  private def user_from_email
    UserQuery.new.email(email.value).first?
  end

  private def passwords_match?; end # check the password
end

# in an action
SignInForm.new(params).submit do |form, user|
  if user
    session[:current_user_id] = user.id.to_s
    redirect to: Dashboard::Show
  else
    render SignInPage, form: form
  end
end
```

paulcsmith:
Closes #55

See #55 for an example

### DIFF
--- a/spec/virtual_form_spec.cr
+++ b/spec/virtual_form_spec.cr
@@ -1,0 +1,47 @@
+require "./spec_helper"
+
+private class TestVirtualForm < LuckyRecord::VirtualForm
+  allow_virtual name : String
+  allow_virtual age : Int32
+
+  def validate
+    validate_required name
+  end
+end
+
+describe LuckyRecord::VirtualForm do
+  it "sets a form_name" do
+    TestVirtualForm.new.form_name.should eq "test_virtual"
+  end
+
+  it "sets up initializers for params and no params" do
+    virtual_form = TestVirtualForm.new
+    virtual_form.name.value.should be_nil
+    virtual_form.name.value = "Megan"
+    virtual_form.name.value.should eq("Megan")
+
+    params = LuckyRecord::Params.new({"name" => "Jordan"})
+    virtual_form = TestVirtualForm.new(params)
+    virtual_form.name.value.should eq("Jordan")
+  end
+
+  it "parses params" do
+    params = LuckyRecord::Params.new({"age" => "45"})
+    virtual_form = TestVirtualForm.new(params)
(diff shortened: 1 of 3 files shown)


jdbann:
Ah great! I was checking against `form.valid?` for a sign in form before this!

paulcsmith:
@jdbann Cool! Do you mind sharing what you had before? If not that's ok, I'm just curious how you got it working without this

jdbann:
```crystal
class SessionForm < User::BaseForm
  allow email
  allow password

  def prepare
    validate_user
  end

  private def validate_user
    unless UserQuery.new.authenticate(email.value.to_s, password.value.to_s).size == 1
      email.add_error "No user exists with this email address and password"
    end
  end
end
```

```crystal
class Sessions::Create < BrowserAction
  action do
    form = SessionForm.new(params)
    if form.valid?
      user = log_in_as_user(form.email.value)
      redirect to: Posts::Index
    else
      render Sessions::NewPage, session_form: form
    end
  end

  private def log_in_as_user(email)
    session["current_user_id"] = UserQuery.new.email.is(email.to_s).first.id
  end
end
```

paulcsmith:
That's pretty clever! BTW you if you aren't chaining things like `lower` you can do `email(email.to_s)`. That is the same as `email.is(email.to_s)`. You may already know that, but I thought I'd let you know just in case :)

jdbann:
Great! I'm still tackling queries but that's a handy tip.

I was going to write up a post about Lucky, exploring it's differences to Rails by walking through a simple signup and authentication process. This felt like it was still a little hacky so I've been holding off but now we have virtual forms!

paulcsmith:
That would be great! If you want to me to take a look at it sometime I'd be happy to do so and give any tips or hints that may be useful

jdbann:
It's still on the list, but depends on some things in `master` so maybe after a release ;)

paulcsmith:
Sounds good :)

Issue #113, opened 2017-11-20
Allow optional belongs_to


paulcsmith:
```crystal
belongs_to category : Category?
```

The code for this goes here: https://github.com/luckyframework/record/blob/master/src/lucky_record/associations.cr

paulcsmith:
Closed by #119

PR #119, opened 2017-11-29
Allow optional belongs_to

mikeeus:
For #113. I used the same type of Union check you used in the `LuckyRecord::Model.field` method to get the model class and pass it to `field`.

LuckyMigrator looks close to done for now so I hope you don't mind if I help out around here. :)

### DIFF
--- a/spec/associations_spec.cr
+++ b/spec/associations_spec.cr
@@ -21,7 +21,7 @@ describe LuckyRecord::Model do
     employee.manager.should eq manager
   end
 
-  it "returns nil for nilable association that doesn't exists" do
+  it "returns nil for nilable association that doesn't exist" do
     employee = EmployeeBox.new.save
     employee.manager.should eq nil
   end
--- a/spec/associations_spec.cr
+++ b/spec/associations_spec.cr
@@ -10,4 +10,19 @@ describe LuckyRecord::Model do
     post.comments.to_a.should eq [comment]
     comment.post.should eq post
   end
+
+  it "gets the related records for nilable association that exists" do
+    manager = ManagerBox.save
+    employee = EmployeeBox.new.manager_id(manager.id).save
+
+    manager = Manager::BaseQuery.new.find(manager.id)
+
+    manager.employees.to_a.should eq [employee]
+    employee.manager.should eq manager
+  end
+
+  it "returns nil for nilable association that doesn't exists" do
+    employee = EmployeeBox.new.save
+    employee.manager.should eq nil
+  end
 end
--- a/spec/associations_spec.cr
+++ b/spec/associations_spec.cr
@@ -3,7 +3,7 @@ require "./spec_helper"
 describe LuckyRecord::Model do
   it "gets the related records" do
     post = PostBox.save
-    comment = CommentBox.new.post_id(post.id).save
+    comment = CommentBox.new.post_id(post.id).save!
 
     post = Post::BaseQuery.new.find(post.id)
 
@@ -13,7 +13,7 @@ describe LuckyRecord::Model do
 
   it "gets the related records for nilable association that exists" do
     manager = ManagerBox.save
-    employee = EmployeeBox.new.manager_id(manager.id).save
+    employee = EmployeeBox.new.manager_id(manager.id).save!
 
     manager = Manager::BaseQuery.new.find(manager.id)
 
--- a/spec/support/base_box.cr
+++ b/spec/support/base_box.cr
@@ -1,12 +1,14 @@
 abstract class BaseBox
   forward_missing_to @record
+  getter form
 
   def initialize
     @record = build_model
   end
 
   macro inherited
-    @record : {{@type.name.gsub(/Box/, "").id}}
+    @record : {{@type.name.gsub(/Box/, "").id}}?
+    form {{ @type.name.gsub(/Box/, "::BaseForm").id }}
   end
 
   def self.build
@@ -14,10 +16,32 @@ abstract class BaseBox
   end
 
   def build
-    @record
+    @record.not_nil!
   end
 
   def build_pair
     [@record, @record]
+  end
+
+  macro form(form_class)
+    @form = {{ form_class }}.new
+  end
+
+  macro method_missing(call)
+    form.{{ call.name }}.value = {{ call.args.first }}
+    form
+  end
+
+  def self.save
+    new.save
+  end
+
+  def save
+    form.save!
+  end
+
+  def self.save_pair
+    new.save
+    new.save
   end
 end
--- a/spec/support/comment.cr
+++ b/spec/support/comment.cr
@@ -3,6 +3,6 @@ require "./post"
 class Comment < LuckyRecord::Model
   table comments do
     field body : String
-    belongs_to post : Post
+    belongs_to post : Post?
   end
 end
--- a/spec/support/comment.cr
+++ b/spec/support/comment.cr
@@ -3,6 +3,6 @@ require "./post"
 class Comment < LuckyRecord::Model
   table comments do
     field body : String
-    belongs_to post : Post?
+    belongs_to post : Post
   end
 end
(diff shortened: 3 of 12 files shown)


spec/support/comment.cr
```diff
@@ -3,6 +3,6 @@ require "./post"
 class Comment < LuckyRecord::Model
   table comments do
     field body : String
-    belongs_to post : Post
+    belongs_to post : Post?
```
paulcsmith: I think this should be changed back because a comment should *always* have a post. Might be worth adding a new table for `Employee` that belongs to `Manager`:

```crystal
class Employee < LuckyRecord::Model
  table :employees do
    belongs_to manager : Manager?
  end
end
```

paulcsmith: There is a spec/associations_spec.cr where you could test this out. One for testing when the associated record is nil, and another for when it is present

mikeeus: Good idea. I created the tables and specs and found an error when the parent model is nil. The `find(id)` method throws `invalid input syntax for integer: ""` so I added an overload `find(id : Nil)` that throws a `RecordNotFoundError` instead.

Is this okay, or should there be different behavior for this case?

src/lucky_record/associations.cr
```diff
@@ -9,9 +9,20 @@ module LuckyRecord::Associations
...
+    {% end %}
+
+    {% if nilable %}
+      field {{ assoc_name.id }}_id : Int32?
+    {% else %}
+      field {{ assoc_name.id }}_id : Int32
+    {% end %}
 
     def {{ assoc_name.id }}
       {{ model }}::BaseQuery.new.find({{ assoc_name.id }}_id)
```
paulcsmith: Instead of adding `find(id: Nil)` and returning `nil`, we should check if the id is present and then make the query:

```crystal
def {{ assoc_name.id }}
  {{ assoc_name.id }}_id.try do |value|
    {{ model }}::BaseQuery.new.find({{ assoc_name.id }}_id)
  end
```

The reason is that `find` should not accept `nil` at all. I'll need to change that later on. I'm planning to rework how types are handled in queries so it might be a bit until I can tackle that

src/lucky_record/associations.cr
```diff
@@ -9,9 +9,20 @@ module LuckyRecord::Associations
...
+    {% else %}
+      {% model = type_declaration.type %}
+      {% nilable = false %}
+    {% end %}
+
+    {% if nilable %}
+      field {{ assoc_name.id }}_id : Int32?
+    {% else %}
+      field {{ assoc_name.id }}_id : Int32
+    {% end %}
```
paulcsmith: Instead of doing two separate lines you could use `if nilable` to show the `?`

```crystal
field {{ assoc_name.id }}_id : Int32{% if nilable %}?{% end %}
```

src/lucky_record/associations.cr
```diff
@@ -9,9 +9,20 @@ module LuckyRecord::Associations
...
-    {% model = type_declaration.type %}
 
-    field {{ assoc_name.id }}_id : Int32
+    {% if type_declaration.type.is_a?(Union) %}
+      {% model = type_declaration.type.types.first %}
+      {% nilable = true %}
+    {% else %}
+      {% model = type_declaration.type %}
+      {% nilable = false %}
+    {% end %}
```
paulcsmith: 👍 Nice!

spec/support/manager_box.cr
```diff
@@ -0,0 +1,26 @@
...
+  def initialize
+    name "Mr. Manager"
+  end
+
+  def self.save
+    new.save
+  end
+
+  def save
+    if form.save
```
paulcsmith: Now you can do `form.save!` and it will raise if it didn't save so you don't need the conditional

spec/support/manager_box.cr
```diff
@@ -0,0 +1,26 @@
+class ManagerBox
+  getter form
+
+  @form = Manager::BaseForm.new
+
+  def initialize
+    name "Mr. Manager"
```
paulcsmith: Haha I love it!

spec/associations_spec.cr
```diff
@@ -10,4 +10,19 @@ describe LuckyRecord::Model do
...
+    manager = ManagerBox.save
+    employee = EmployeeBox.new.manager_id(manager.id).save
+
+    manager = Manager::BaseQuery.new.find(manager.id)
+
+    manager.employees.to_a.should eq [employee]
+    employee.manager.should eq manager
+  end
+
+  it "returns nil for nilable association that doesn't exists" do
```
paulcsmith: exists -> exist

spec/support/employee_box.cr
```diff
@@ -0,0 +1,26 @@
+class EmployeeBox
+  getter form
+
+  @form = Employee::BaseForm.new
```
paulcsmith: If you're up for it, maybe you could extract a `BaseBox` class? Something like this: https://gist.github.com/paulcsmith/c5da6e8888d1a50f13fb6d9acc37a6a9#file-base_box-cr

src/lucky_record/associations.cr
mikeeus: Wow I didn't know I could do that

src/lucky_record/associations.cr
```diff
@@ -9,12 +9,21 @@ module LuckyRecord::Associations
...
+
+    field {{ assoc_name.id }}_id : Int32{% if nilable %}?{% end %}
 
     def {{ assoc_name.id }}
-      {{ model }}::BaseQuery.new.find({{ assoc_name.id }}_id)
+      {{ assoc_name.id }}_id.try do |value|
+        {{ model }}::BaseQuery.new.find({{ assoc_name.id }}_id)
```
paulcsmith: Oh and I thin this should use the `value` in the `find`. Just to make sure it uses the not nil version: `find(value)`. That was my bad. I think I did it wrong in the snippet I posted earlier

mikeeus:
K I refactored the forms into BaseBox and the specs are passing. I made @record nilable because it's not set when the form is created with `initialize`, then added a `not_nil!` call on `build`.

spec/support/base_box.cr
```diff
@@ -1,23 +1,47 @@
...
   def build
-    @record
+    @record.not_nil!
   end
 
   def build_pair
     [@record, @record]
   end
+
+  macro form(form_class)
```
paulcsmith: Very nice!

spec/support/base_box.cr
```diff
@@ -1,23 +1,47 @@
...
   forward_missing_to @record
+  getter form
 
   def initialize
     @record = build_model
   end
 
   macro inherited
-    @record : {{@type.name.gsub(/Box/, "").id}}
+    @record : {{@type.name.gsub(/Box/, "").id}}?
```
paulcsmith: This is ok 👍

spec/associations_spec.cr
```diff
@@ -3,11 +3,26 @@ require "./spec_helper"
...
+    manager = ManagerBox.save
+    employee = EmployeeBox.new.manager_id(manager.id).save!
+
+    manager = Manager::BaseQuery.new.find(manager.id)
+
+    manager.employees.to_a.should eq [employee]
+    employee.manager.should eq manager
+  end
+
+  it "returns nil for nilable association that doesn't exist" do
```
paulcsmith: doesn't -> don't

mikeeus: I think it's "doesn't" since "association" is singular. `associations that don't exist` vs `association that doesn't exist`

src/lucky_record/associations.cr
mikeeus: Ah yes, you mean like this?
```crystal
{{ assoc_name.id }}_id.try do |value|
  {{ model }}::BaseQuery.new.find(value)
end
```

paulcsmith: Yep!

spec/associations_spec.cr
paulcsmith: Derp. I misread. Carry on! :)

paulcsmith:
Wonderful job 💯

Issue #111, opened 2017-11-19
Query with “in”


paulcsmith:
`PostQuery.new.id.in([1, 2, 3])`

paulcsmith:
Closed by #120

PR #120, opened 2017-11-29
Add #in query using '= ANY' and '!= ALL'

mikeeus:
Here's my first go at #111.

I used `= ANY` because it expects an array whereas [the IN comparison expects a comma separated list of scalar expressions](https://www.postgresql.org/docs/9.1/static/functions-comparisons.html). The pg driver inserts an Array(Int32) directly like `{1,2,3}` and I couldn't separate the elements into scalar values without using multiple placeholders which i think is unnecessary.

There may be a way to do it with `IN`, but I don't know how and I don't think there are any drawbacks of using `ANY`.

Anyway, great work on this. I can't wait to use it on a project when I get the time.

### DIFF
--- a/src/lucky_record/charms/int32_extensions.cr
+++ b/src/lucky_record/charms/int32_extensions.cr
@@ -17,8 +17,16 @@ struct Int32
       SuccessfulCast(Int32).new(value)
     end
 
+    def self.parse(value : Array(Int32))
+      SuccessfulCast(Array(Int32)).new(value)
+    end
+
     def self.to_db(value : Int32)
       value.to_s
+    end
+
+    def to_db(value : Array(Int32))
+      value
     end
 
     class Criteria(T, V) < LuckyRecord::Criteria(T, V)
--- a/src/lucky_record/charms/int32_extensions.cr
+++ b/src/lucky_record/charms/int32_extensions.cr
@@ -17,16 +17,8 @@ struct Int32
       SuccessfulCast(Int32).new(value)
     end
 
-    def self.parse(value : Array(Int32))
-      SuccessfulCast(Array(Int32)).new(value)
-    end
-
     def self.to_db(value : Int32)
       value.to_s
-    end
-
-    def to_db(value : Array(Int32))
-      value
     end
 
     class Criteria(T, V) < LuckyRecord::Criteria(T, V)
--- a/src/lucky_record/criteria.cr
+++ b/src/lucky_record/criteria.cr
@@ -50,6 +50,10 @@ class LuckyRecord::Criteria(T, V)
     add_clause(LuckyRecord::Where::LessThanOrEqualTo.new(column, V::Lucky.to_db!(value)))
   end
 
+  def in(value)
+    add_clause(LuckyRecord::Where::In.new(column, V::Lucky.to_db!(value)))
+  end
+
   private def add_clause(sql_clause)
     sql_clause = build_sql_clause(sql_clause)
     rows.query.where(sql_clause)
--- a/src/lucky_record/criteria.cr
+++ b/src/lucky_record/criteria.cr
@@ -50,8 +50,9 @@ class LuckyRecord::Criteria(T, V)
     add_clause(LuckyRecord::Where::LessThanOrEqualTo.new(column, V::Lucky.to_db!(value)))
   end
 
-  def in(value)
-    add_clause(LuckyRecord::Where::In.new(column, V::Lucky.to_db!(value)))
+  def in(values)
+    values = values.map { |value| V::Lucky.to_db!(value) }
+    add_clause(LuckyRecord::Where::In.new(column, values))
   end
 
   private def add_clause(sql_clause)
--- a/src/lucky_record/where.cr
+++ b/src/lucky_record/where.cr
@@ -2,14 +2,18 @@ module LuckyRecord::Where
   abstract class SqlClause
     getter :column, :value
 
-    def initialize(@column : Symbol | String, @value : String)
+    def initialize(@column : Symbol | String, @value : String | Array(String) | Array(Int32))
     end
 
     abstract def operator : String
     abstract def negated : SqlClause
 
     def prepare(prepared_statement_placeholder : String)
-      "#{column} #{operator} #{prepared_statement_placeholder}"
+      if ["= ANY", "!= ALL"].includes? operator
+        "#{column} #{operator} (#{prepared_statement_placeholder})"
+      else
+        "#{column} #{operator} #{prepared_statement_placeholder}"
+      end
     end
   end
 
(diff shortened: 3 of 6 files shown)


src/lucky_record/where.cr
```diff
@@ -112,4 +116,24 @@ module LuckyRecord::Where
...
+      NotIn.new(@column, @value)
+    end
+  end
+
+  class NotIn < SqlClause
+    def operator
+      "!= ALL"
+    end
+
+    def negated : In
```
hanneskaeufler: We could add two expectations in the `where_spec.cr` to make sure the negated part is as expected.

src/lucky_record/where.cr
```diff
@@ -2,14 +2,18 @@ module LuckyRecord::Where
...
-    def initialize(@column : Symbol | String, @value : String)
+    def initialize(@column : Symbol | String, @value : String | Array(String) | Array(Int32))
     end
 
     abstract def operator : String
     abstract def negated : SqlClause
 
     def prepare(prepared_statement_placeholder : String)
-      "#{column} #{operator} #{prepared_statement_placeholder}"
+      if ["= ANY", "!= ALL"].includes? operator
```
hanneskaeufler: If possible I guess we would like to not repeat those string literals here (and maybe get rid of the conditional?). I was going to suggest something like

```
operator = "ANY"

prepare = {
  "ANY" => ->(statement : String) { "(#{statement})" },
  "NOT ANY" => ->(statement : String) { "(#{statement})" }
  }.fetch(operator, ->(statement : String) { statement })


puts "#{prepare.call("$1")}"
puts "#{prepare.call("$3")}"
```

To replace the conditional by a lookup table but seeing it now its both more code and harder to understand on first sight haha. But still, removing the repeated operators could be a good idea?

src/lucky_record/where.cr
mikeeus: Oh I didn't see those specs, thanks!

paulcsmith:
I'll have to review this later, but I'm adding these resources that will be helpful to later me and possibly others:

* https://stackoverflow.com/questions/34627026/in-vs-any-operator-in-postgresql/34627688
* https://stackoverflow.com/questions/31191507/how-to-use-any-instead-of-in-in-a-where-clause-with-rails/31192557#31192557

Thanks for tackling this! This will be necessary for preloading. I'm excited :)

src/lucky_record/where.cr
mikeeus: Yeah I agree removing repetition would be good if we end up with more conditions, but I think this is okay for now right?

hanneskaeufler: Sure, not a show stopper imho :) Alternatively we could make `operator` a function that accepts a string?

so e.g. for `Equal` this would get:
```crystal
def operator(placeholder : String)
    "= #{placeholder}"
end
```

and e.g. for  `In`
```crystaö
def operator(placeholder : String)
  "ANY (#{placeholder})"
end
```

Just throwing this out there :)

paulcsmith: Another option is to define a `prepare` method on `SqlClause` and then you can override it in `In` and `NotIn`. That would probably be simpler and would keep the customization in the class that needs it. LMK if you need me to expand on this or if it is not clear

src/lucky_record/criteria.cr
```diff
@@ -50,6 +50,10 @@ class LuckyRecord::Criteria(T, V)
     add_clause(LuckyRecord::Where::LessThanOrEqualTo.new(column, V::Lucky.to_db!(value)))
   end
 
+  def in(value)
+    add_clause(LuckyRecord::Where::In.new(column, V::Lucky.to_db!(value)))
```
paulcsmith: Rather than adding parsing to all the types, you could probably remove the new parse and to_db methods and instead map the values:

```
def in(values)
  values = values.map { |value| V::Lucky.to_db!(value) }
  add_clause(LuckyRecord::Where::In.new(column, values))
end
```

Then you'll have an array of values. I'm not if this will require changes elsewhere in LuckyRecord, but I hope not *fingers crossed*

mikeeus:
Thanks for the links @paulcsmith. I took a look at them, and I did more searching but I can't find a way to convert an array into a set so it can be used by `IN`. I'll keep my eyes open and let you know if I find anything.

src/lucky_record/where.cr
mikeeus: K added the `prepare` method to `In` and `NotIn`, that cleaned it up nicely

src/lucky_record/criteria.cr
paulcsmith: @mikeeus I tried this out locally and specs seemed to pass. Here's a commit on a test branch that you can check out. I *think* this should work: https://github.com/luckyframework/lucky_record/commit/4e03f4e23926e5b60d9072b44e323d9dfe7a5967

mikeeus: Nice, this works very well

src/lucky_record/charms/int32_extensions.cr
```diff
@@ -18,7 +18,7 @@ struct Int32
     end
 
     def self.to_db(value : Int32)
-      value.to_s
+      value
```
mikeeus: This change broke a lot of tests that expected Int32 to be a string.

src/lucky_record/criteria.cr
```diff
@@ -50,7 +50,7 @@ class LuckyRecord::Criteria(T, V)
     add_clause(LuckyRecord::Where::LessThanOrEqualTo.new(column, V::Lucky.to_db!(value)))
   end
 
-  def in(values)
+  def in(values : Array(Int32) | Array(String)) : T
```
mikeeus: I added this to ensure that the `type_cast` method works.

src/lucky_record/where.cr
```diff
@@ -115,29 +123,29 @@ module LuckyRecord::Where
...
+      "IN"
     end
 
     def negated : NotIn
       NotIn.new(@column, @value)
     end
 
     def prepare(prepared_statement_placeholder : String)
-      "#{column} #{operator} (#{prepared_statement_placeholder})"
+      "#{column} #{operator} (SELECT * FROM unnest(#{prepared_statement_placeholder}#{type_cast}))"
```
mikeeus: I finally found a way to use an `IN` statement with an array, but the array needs to be casted to a specific type. So I added a type_cast method and restricted `.in()` calls to `Array(Int32)` and `Array(String)`.

src/lucky_record/charms/int32_extensions.cr
paulcsmith: You’re probably right but I think it’s be better to handle that separately. Could you open an issue about removing the to_s?

mikeeus:
K I opened an issue and added a PR for the Int32 thing.

I agree on using `ANY`, I think that's the way to go. And no problem, it was a learning experience if nothing else. I don't have time now, but I'll work on it tomorrow.

mikeeus:
Alright, I reset it to the `map values for in` commit and rebased it.

paulcsmith:
Awesome! Thanks a ton :D

PR #145, opened 2017-12-05
Do not try to save virtual fields to db.

oneiros:
First of all: I am super excited about lucky. Thanks for sharing!

Following the guide for adding virtual fields to forms, I encountered a small bug: When creating records lucky tried to save the virtual fields as well as the actual database fields.

This PR includes a very tiny fix that worked for me. Of course, I cannot be 100% sure that this does not have any side-effects I did not consider.

Also, the added test might be a bit clumsy. Let me know if there is anything I can improve.

### DIFF
--- a/spec/allow_virtual_spec.cr
+++ b/spec/allow_virtual_spec.cr
@@ -58,6 +58,13 @@ describe "allow_virtual in forms" do
     form.setup_required_database_fields
     form.valid?.should be_false
   end
+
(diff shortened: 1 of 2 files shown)


paulcsmith:
I’m so glad to hear you’re excited about Lucky!

Thanks so much for catching this. The tests and the fix look perfect. 👍 

I really appreciate your help

PR #149, opened 2017-12-10
Return nil when value is a blank string

### LINKED ISSUE #148: Field#value should return `nil` if the string is blank

mikeeus:
Closes #148

### DIFF
--- a/spec/form_spec.cr
+++ b/spec/form_spec.cr
@@ -203,6 +203,14 @@ describe "LuckyRecord::Form" do
       form.name.value.should eq "Joe"
       form.name.errors.should eq ["wrong"]
     end
+
+    it "Field#value returns nil on empty strings" do
+      empty_string = LuckyRecord::Field.new(:blank, nil, " ", "test_form")
+      empty_string.value.should be_nil
+
+      empty_array = LuckyRecord::Field.new(:empty_array, nil, [] of String, "test_form")
+      empty_array.value.should_not be_nil
+    end
   end
 
   describe ".save" do
--- a/spec/form_spec.cr
+++ b/spec/form_spec.cr
@@ -203,14 +203,6 @@ describe "LuckyRecord::Form" do
       form.name.value.should eq "Joe"
       form.name.errors.should eq ["wrong"]
     end
-
-    it "Field#value returns nil on empty strings" do
-      empty_string = LuckyRecord::Field.new(:blank, nil, " ", "test_form")
-      empty_string.value.should be_nil
-
-      empty_array = LuckyRecord::Field.new(:empty_array, nil, [] of String, "test_form")
-      empty_array.value.should_not be_nil
-    end
   end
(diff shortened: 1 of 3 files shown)


src/lucky_record/field.cr
```diff
@@ -20,11 +20,19 @@ class LuckyRecord::Field(T)
...
     add_error(message_string)
   end
-  
+
   def errors
     @errors.uniq
   end
 
+  def value
+    if @value == ""
```
paulcsmith: If you use “@value.blank?” It will also catch strings that are all spaces. You’d also have to check that it is a String though so it doesn’t catch other objects that implement blank. So maybe something like: @value.is_a? && @value.blank?

mikeeus: I didn't think of those cases, I'll update it.

mikeeus:
I updated ` def value` to check for blank string but it's giving me this error:
```crystal
in src/lucky_record/field.cr:29: undefined method 'blank?' for Array(String)

    if @value.is_a?(String) && @value.blank?
                                      ^~~~~~
```

`@value.is_a?(String)` should fix this error but it's still raising.

paulcsmith:
@mikeeus Hmm that's tricky. I think it is because `@value` could change because it is an instance variable. Could you try first assinging to a local variable:

```crystal
value = @value
if value.is_a?(String)....
```

I *think* that should work

mikeeus:
It worked! I think it's because it is `Generic`, because I tried the same with `@param` which is `String?` and it worked. Either it's good to know, thanks!

spec/form_spec.cr
```diff
@@ -203,6 +203,14 @@ describe "LuckyRecord::Form" do
       form.name.value.should eq "Joe"
       form.name.errors.should eq ["wrong"]
     end
+
+    it "Field#value returns nil on empty strings" do
+      empty_string = LuckyRecord::Field.new(:blank, nil, " ", "test_form")
+      empty_string.value.should be_nil
+
+      empty_array = LuckyRecord::Field.new(:empty_array, nil, [] of String, "test_form")
```
paulcsmith: Could you use names arguments so it’s a bit easier to see what they each option is? E.g. Field.new(name: :empty_array, ...)

spec/form_spec.cr
```diff
@@ -203,6 +203,14 @@ describe "LuckyRecord::Form" do
       form.name.value.should eq "Joe"
       form.name.errors.should eq ["wrong"]
     end
+
+    it "Field#value returns nil on empty strings" do
```
paulcsmith: Maybe extract a field_spec.cr spec so any future field specific tests can go there. What do you think?

mikeeus: Sure, I was thinking the same but I didn't since there were only two specs.

paulcsmith:
Thanks!

Issue #143, opened 2017-12-04
Raise LuckyRecord::RecordNotFoundError for `first`, `last` etc. when a record is not found


paulcsmith:
Right now it raises for `find` but not for `first`, `last` etc.

Similar to 
https://github.com/luckyframework/lucky_record/blob/acdc00a3237e186f5a391123bca189149aa9453c/src/lucky_record/queryable.cr#L39

This requires:

* Making sure `first` and `last` raise the error: `results.first? || raise RecordNotFoundError.new(...the_args)`

paulcsmith:
Closed by #156 and #150

PR #150, opened 2017-12-10
Raise RecordNotFound for Queryable#first

mikeeus:
For #143.

There is no `#last` method, should I make an issue and add it?

I had to add an overload to `RecordNotFound` to handle generating an error message for query methods like `first`, but the attribute is still called `id` which may be misleading. Is this okay, or is there a better way to handle it?

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -20,6 +20,12 @@ describe LuckyRecord::Query do
       insert_a_user
       UserQuery.new.first.name.should eq "Paul"
     end
+
+    it "raises RecordNotFound if no record is found" do
+      expect_raises(LuckyRecord::RecordNotFoundError, "") do
+        UserQuery.new.first
+      end
+    end
   end
 
   describe "#find" do
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -22,7 +22,7 @@ describe LuckyRecord::Query do
     end
 
     it "raises RecordNotFound if no record is found" do
-      expect_raises(LuckyRecord::RecordNotFoundError, "") do
+      expect_raises(LuckyRecord::RecordNotFoundError) do
         UserQuery.new.first
       end
     end
@@ -37,13 +37,13 @@ describe LuckyRecord::Query do
     end
 
     it "raises RecordNotFound if no record is found with the given id (Int32)" do
-      expect_raises(LuckyRecord::RecordNotFoundError, "") do
+      expect_raises(LuckyRecord::RecordNotFoundError) do
         UserQuery.new.find(1)
       end
     end
 
     it "raises RecordNotFound if no record is found with the given id (String)" do
-      expect_raises(LuckyRecord::RecordNotFoundError, "") do
+      expect_raises(LuckyRecord::RecordNotFoundError) do
         UserQuery.new.find("1")
       end
     end
--- a/src/lucky_record/errors.cr
+++ b/src/lucky_record/errors.cr
@@ -5,7 +5,7 @@ module LuckyRecord
   class LuckyRecordError < Exception
   end
 
-  # Raised when a record could not be find
+  # Raised when a record could not be found
   class RecordNotFoundError < LuckyRecordError
     def initialize(@model : Symbol, @id : String)
       super "Could not find #{model} with id of #{id}"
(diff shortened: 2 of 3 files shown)


src/lucky_record/errors.cr
```diff
@@ -12,6 +14,13 @@ module LuckyRecord
     end
   end
 
+  # Raised when Lucky Record cannot find a record by a query_method like #first or #last
+  class RecordNotFoundError < LuckyRecordError
+    def initialize(@model : Symbol, @id : Symbol)
```
paulcsmith: You could put this overloaded method in the existing class so they don’t get split up. The comment could change to something like: “raided when a record could not be found”

src/lucky_record/errors.cr
```diff
@@ -12,6 +14,13 @@ module LuckyRecord
...
+      super "Could not find #{id} #{LuckyInflector::Inflector.singularize(model)}"
```
paulcsmith: Maybe this could be simplified to: “could not find first record in #{model}”. You could then pass in “first” so later you could also pass in “last” or whatever else we decide to add later.

src/lucky_record/queryable.cr
```diff
@@ -42,6 +42,8 @@ module LuckyRecord::Queryable(T)
   def first
     query.limit(1)
     exec_query.first
```
paulcsmith: Instead of rescuing you could do: “first? || raise ...”

paulcsmith:
Oh and yes, I’d you could add an issue for adding last that would be great!

src/lucky_record/errors.cr
mikeeus: Of course, I'm not sure why I made it a separate class.

What does raided mean in this context?

paulcsmith: Oops! That was a typo. I meant “raises”

spec/query_spec.cr
```diff
@@ -20,6 +20,12 @@ describe LuckyRecord::Query do
       insert_a_user
       UserQuery.new.first.name.should eq "Paul"
     end
+
+    it "raises RecordNotFound if no record is found" do
+      expect_raises(LuckyRecord::RecordNotFoundError, "") do
```
paulcsmith: I believe you can remove the second option from this

src/lucky_record/errors.cr
```diff
@@ -5,11 +5,15 @@ module LuckyRecord
   class LuckyRecordError < Exception
   end
 
-  # Raised when Lucky Record cannot find a record by given id
+  # Raised when a record could not be find
```
paulcsmith: Small typo. Find -> found

src/lucky_record/errors.cr
```diff
@@ -5,11 +5,15 @@ module LuckyRecord
...
   class RecordNotFoundError < LuckyRecordError
     def initialize(@model : Symbol, @id : String)
       super "Could not find #{model} with id of #{id}"
     end
+
+    def initialize(@model : Symbol, @id : Symbol)
```
paulcsmith: Instead of calling it id, what do you think about “query”? Not perfect but maybe a bit more clear than id?

mikeeus: Yeah it's a bit clearer.

mikeeus: Okay, so if I rename the overload to `query` then I need to make the first `id` argument nilable.

```bash
The instance variable '@id' is initialized in other 'initialize' methods,
and by not initializing it here it's not clear if the variable is supposed
to be nilable or if this is a mistake.

To fix this error, either assign nil to it here:

  @id = nil

Or declare it as nilable outside at the type level:

  @id : (String)?

    def initialize(@model : Symbol, @query : Symbol)

```

paulcsmith: That's a good error and it highlights that we probably don't need instance variables. You can use local variables instead and the issue should go away

mikeeus: Okay so its because the instance variable is set on the class and needs to be typed, got it

paulcsmith:
Looks great!

src/lucky_record/errors.cr
paulcsmith: Kind of. The problem was that `@id` was not nilable, but if you called the `def initialize(@model : Symbol, @query : Symbol)` then `@id` would be `nil`. So you it was saying you either needed to set the variable or make it nilable with String?

By using local vars there is no issue because not all initializers need to set them

PR #151, opened 2017-12-10
Add join support

### LINKED ISSUE #128: Add join support
> This will be fairly difficult to tackle, but I'm happy to help. Feel free to reach out in https://gitter.im/luckyframework/Lobby for help

When you define an association with has_many or belongs_to it should add a number of fields to the BaseQuery 

```crystal
class Comment < BaseModel
  table :comments do
    belongs_to post : Post
  end
end

Comment::BaseQuery.new.join_post 
# or
.inner_join_post
.left_join_post
```

This should also work for has_many associations. This isn't that helpful on its own, but it will be required for #61 and #59 

This will require:

* Adding a `@joins` to the QueryBuilder https://github.com/luckyframework/lucky_record/blob/8607673e9d6fe43382ee06eb55ba7fdbecad83f4/src/lucky_record/query_builder.cr#L5 to hold all the join clauses
* Add the join clauses the sql_statements method. This wll work similarly to how the `where_sql` is generated. https://github.com/luckyframework/lucky_record/blob/8607673e9d6fe43382ee06eb55ba7fdbecad83f4/src/lucky_record/query_builder.cr#L51
* Keep track of associations that were added, similar to what was done with `FIELDS` https://github.com/luckyframework/lucky_record/blob/8607673e9d6fe43382ee06eb55ba7fdbecad83f4/src/lucky_record/model.cr#L5
* Pass them to the base query template: https://github.com/luckyframework/lucky_record/blob/8607673e9d6fe43382ee06eb55ba7fdbecad83f4/src/lucky_record/model.cr#L64-L66
* Create methods in the BaseQueryTemplate for each of the associations: https://github.com/luckyframework/lucky_record/blob/master/src/lucky_record/base_query_template.cr

mikeeus:
Closes #128.

This one was really frustrating. I couldn't get LuckyInflector to work inside a macro so I could pluralize the belongs_to table reference.


spec/query_spec.cr
```diff
@@ -177,6 +177,28 @@ describe LuckyRecord::Query do
...
+
+  # how should I test these methods?
+  pending "#joins methods for associations" do
+    it "inner joins on belongs to" do
+      insert_a_post
+      post = Post::BaseQuery.new.first
+      insert_a_comment(post.id)
+
+      result = Comment::BaseQuery.new.join_posts.first # creates join_post at the moment
+      result.post.should eq post
```
mikeeus: How should I test that this works. Would I need to add a `post` field to the `Comment` and load the post there?

src/lucky_record/associations.cr
```diff
@@ -9,6 +14,9 @@ module LuckyRecord::Associations
 
   macro belongs_to(type_declaration)
     {% assoc_name = type_declaration.var }
+    {% LuckyInflector::Inflector.pluralize assoc_name %} # Can't get this to work...
+
+    association {{ assoc_name }}, :id
```
mikeeus: This is what really frustrated me. `LuckyInflector::Inflector` is a `TypeNode` so I can't call it's methods, only list them through `class.methods`. But I need to somehow pluralize the `belongs_to` association for the generated methods like `join_posts` and for the join statement itself.

Any thoughts?

paulcsmith: Ah yeah this is a bit tricky. You can’t run any runtime code in a macro call. There some ways to do it, but i think there is a better option in this case. 

In the model we set @@table_name. We can also set a TABLE_NAME constant. This is because constants can be accessed in macros by calling .constant(:THE_CONSTANT_NAME)

So, if we set the TABLE_NAME constant we could get the associations table name in this macro like this: `type_declaration.type.constant(:TABLE_NAME)`

Does that make sense?

spec/query_spec.cr
paulcsmith: I think for now a simple test like this is good, but it should also test the generated SQL. Until there is support for querying associations it’ll be hard to test without testing the generated SQL directly.

src/lucky_record/associations.cr
mikeeus: That's perfect. I dug through the `TypeNode` documentation and it never occurred to me to use a constant. Thanks for this, I'm sure I'll find it useful in the future.

spec/query_spec.cr
```diff
@@ -189,3 +217,22 @@ private def insert_a_user(name = "Paul", age = 34)
...
+
+private def insert_a_comment(post_id, body = "Slick as Ruby")
+  LuckyRecord::Repo.run do |db|
+    db.exec "INSERT INTO comments(post_id, body, created_at, updated_at) VALUES ($1, $2, $3, $4)",
+      post_id,
+      body,
+      Time.now,
+      Time.now
+  end
+end
```
paulcsmith: You can now use "boxes" which will handle most of this for you: https://github.com/luckyframework/lucky_record/tree/master/spec/support/boxes

See it in action here: https://github.com/luckyframework/lucky_record/pull/135/files#diff-5085ddc121b47697c6b81979a09f077fR19

src/lucky_record/associations.cr
```diff
@@ -1,6 +1,11 @@
+require "lucky_inflector"
```
paulcsmith: I think this can be removed now

src/lucky_record/associations.cr
```diff
@@ -20,6 +25,8 @@ module LuckyRecord::Associations
 
     field {{ assoc_name.id }}_id : Int32{% if nilable %}?{% end %}
 
+    association {{ model.resolve.constant(:TABLE_NAME).id }}, :id
```
paulcsmith: Could you use named args for these? `association table_name: {{ }}, foreign_key: :id` so it's a bit more clear what is going on

spec/query_spec.cr
mikeeus: I forgot about boxes, they're very convenient.

src/lucky_record/associations.cr
mikeeus: Sure thing, that's much cleaner

spec/query_spec.cr
```diff
@@ -186,9 +186,8 @@ describe LuckyRecord::Query do
 
   describe "#join methods for associations" do
     it "inner join on belongs to" do
-      insert_a_post
-      post = Post::BaseQuery.new.first
-      insert_a_comment(post.id)
+      post = PostBox.new.save
```
mikeeus: I tried to use `PostBox.new.save!` here and it raised:
```
in spec/query_spec.cr:183: instantiating 'PostBox#save!()'

      post = PostBox.new.save!
                         ^~~~~ 
```
But it works fine for `CommentBox.new.post_id(post.id).save!` since `post_id()` returns a form, and under the hood `PostBox.save` does `form.save!`.

Is there are reason why there isn't a `save!` method on the Box itself, because I think this can cause confusion.

paulcsmith: You are totally right. Could you open an issue about that so usability can be improved?

paulcsmith:
This is really nice 👍

spec/query_spec.cr
paulcsmith: Actually I can do it :) Thanks for bringing it up

Issue #154, opened 2017-12-12
Make `Box.save` method consistent


paulcsmith:
https://github.com/luckyframework/lucky_record/pull/151#discussion_r156393551

paulcsmith:
Closed by https://github.com/luckyframework/lucky_record/pull/159

PR #159, opened 2017-12-15
Use UserBox instead of custom factory method

jsteiner:
Change `LuckyRecord::Box` to return `self`, so setters can be chainable.


paulcsmith:
Closes #159 
Closes #154

PR #163, opened 2017-12-16
Ensure correct field mappings

### LINKED ISSUE #127: Add methods for raising if there are incorrect field mappings
Run `LuckyRecord::Model.ensure_correct_field_mappings!`

Raises if:

* There is a field defined in the model doesn't have a matching column on the table
* The table is missing
* The field is marked as nilable with `?` but the column on the table is NON NULL
* A field is marked as non-nilable (no `?`), but the column allows NULL

Ideally it should use https://crystal-lang.org/api/0.23.1/Levenshtein.html to check for similar named tables/field and suggest corrections. You can see how this is used for missing assets in Lucky: https://github.com/luckyframework/lucky/blob/e6c5efc578b9753816985db4fc44fae589b7573f/src/run_macros/missing_asset.cr#L10

It should also give helpful suggestions when it raises:

```
<FIELD NAME> is marked as nilable (nickname : String?), but the database column does not allow nils.

Try this...
   
  * Mark the field as non-nilable in your model: field nickname : String
  * Or, change the column in a migration to allow nils: make_optional :users, :nickname
```

It does *not* need to find and display all problems at once. So if there is a missing table *and* a model has a missing field, just display the error for a missing table. Then once the table is added and they run it again show the error for a missing field.

mikeeus:
Closes #127.

I was busy this week but now I'm free to crush this issue today. :)

I moved the `select_rows` and `table_name` methods from the `DatabaseCleaner` to `Repo` since they're  useful in other places. Besides that all the changes are going to be in `Model`.

I'm thinking of adding `ensure_optional_fields_are_nilable!` and `ensure_required_fields_are_not_nilable!`, then running through them one at a time.


spec/model_spec.cr
```diff
@@ -9,6 +9,29 @@ private class QueryMe < LuckyRecord::Model
   end
 end
 
+# USERS
+# id serial PRIMARY KEY,
+# name text NOT NULL,
+# created_at timestamp NOT NULL,
+# updated_at timestamp NOT NULL,
+# age int NOT NULL,
+# nickname text,
```
mikeeus: These comments are just for my reference, I'll remove them once the tests pass.

mikeeus:
Alright, here's my go at it. Although the `ensure_fields_match_columns` method is big, I think it's pretty easy to follow. Let me know what you think :)

src/lucky_record/model.cr
```diff
@@ -103,4 +107,89 @@ class LuckyRecord::Model
...
+    Try this...
+
+      * Mark #{field[:name]} as non-nilable in your model: #{field[:name]} : #{field[:type]}
+      * Or, change the column in a migration to allow nils: make_optional :users, :#{field[:name]}
+    ERROR
+  end
+
+  private def required_field_error(field)
+    <<-ERROR
+    #{field[:name]} is marked as required (#{field[:name]} : #{field[:type]}), but the database column it not.
```
hanneskaeufler: `but the database column isn't`?

src/lucky_record/model.cr
```diff
@@ -103,4 +107,89 @@ class LuckyRecord::Model
...
+
+    Try this...
+
+      * Mark #{field[:name]} as nilable in your model: #{field[:name]} : #{field[:type]}?
+      * Or, change the column in a migration to allow nils: make_optional :#{@@table_name}, :#{field[:name]}
```
hanneskaeufler: Does it make sense to recommend making the db column optional when the field is required?

spec/model_spec.cr
```diff
@@ -9,6 +9,29 @@ private class QueryMe < LuckyRecord::Model
...
+end
+
+private class RequiredFieldOnOptionalColumn < LuckyRecord::Model
+  table users do
+    field nickname : String
+  end
+end
+
+private class MissingTable < LuckyRecord::Model
+  table uusers do
```
hanneskaeufler: `uusers` looks like an undeliberate typo :) maybe name it `intentionally_missing_table` haha :)

spec/model_spec.cr
```diff
@@ -97,4 +120,48 @@ describe LuckyRecord::Model do
...
+    end
+
+    # field defined in model without a matching column in table
+    it "raises on fields with missing columns" do
+      now = Time.now
+      user = MissingColumn.new id: 1,
+        created_at: now,
+        updated_at: now,
+        missing: "missing"
+      expect_raises Exception, "The table users does not have a 'missing' column. Make sure you've added it to the migration." do
```
hanneskaeufler: Its not relevant to which migration you actually add a field, is it? The `to the migration` I find slightly confusing. To which migration? Maybe `added it in one of your apps migrations` or something?

spec/model_spec.cr
mikeeus: lol good point, I'll change that

src/lucky_record/model.cr
mikeeus: Good catch

src/lucky_record/model.cr
mikeeus: I think the message is intended to offer both possible solutions because the mistake could be either

spec/model_spec.cr
mikeeus: Sure, how about `Make sure you've added it to a migration`

spec/model_spec.cr
mikeeus: Ah I just realized, the reason I made it similar to users is because it checks for similarity with `Levenshtein` to suggest a correction

src/lucky_record/model.cr
hanneskaeufler: I'm still not sure I understand. `#{field[:name]} is marked as required`. Is allowing nils for the column really a possible solution? Wouldn't we need to make the db column not-nilable for this case?

mikeeus: The user could have written the migrations first and marked a column `NOT NULL`, then made a mistake on the model field marking it as nilable`?`. Or the user might change the model without realizing the database column is still the same.

I think the error is intended to suggest solutions to both cases, or at least that's my understanding

spec/model_spec.cr
```diff
@@ -9,6 +9,29 @@ private class QueryMe < LuckyRecord::Model
...
+  end
+end
+
+private class RequiredFieldOnOptionalColumn < LuckyRecord::Model
+  table users do
+    field nickname : String
+  end
+end
+
+private class MissingTable < LuckyRecord::Model
```
paulcsmith: Maybe `MissingButSimilarlyNamedTable`? So it's a bit more clear that it is purposely slightly misspelled?

spec/model_spec.cr
```diff
@@ -9,6 +9,29 @@ private class QueryMe < LuckyRecord::Model
...
+  end
+end
+
+private class OptionalFieldOnRequiredColumn < LuckyRecord::Model
+  table users do
+    field name : String?
+  end
+end
+
+private class RequiredFieldOnOptionalColumn < LuckyRecord::Model
```
paulcsmith: Nice name 👍

spec/model_spec.cr
```diff
@@ -97,4 +120,48 @@ describe LuckyRecord::Model do
...
   end
+
+  describe ".ensure_correct_field_mappings" do
+    # table is missing
+    it "raises on missing table" do
+      missing_table = MissingTable.new(1, Time.new, Time.new)
+      expect_raises Exception, "The table uusers was not found, did you mean users?" do
+        missing_table.ensure_correct_field_mappings!
+      end
+    end
```
paulcsmith: I think another case for when the missing table doesn't have a similarly named table would also be good.

spec/model_spec.cr
```diff
@@ -97,4 +120,48 @@ describe LuckyRecord::Model do
       QueryMe.column_names.should eq [:id, :created_at, :updated_at, :email, :age]
     end
   end
+
+  describe ".ensure_correct_field_mappings" do
+    # table is missing
+    it "raises on missing table" do
+      missing_table = MissingTable.new(1, Time.new, Time.new)
+      expect_raises Exception, "The table uusers was not found, did you mean users?" do
```
paulcsmith: I think switching `,` to a period `.` would be more correct. Also maybe adding the yellow color to `Did you mean 'users'?` would be cool:

```crystal
"The table 'uusers' was not found. #{"Did you mean users?".colorize("yellow")}"
```

spec/model_spec.cr
```diff
@@ -97,4 +120,48 @@ describe LuckyRecord::Model do
       QueryMe.column_names.should eq [:id, :created_at, :updated_at, :email, :age]
     end
   end
+
+  describe ".ensure_correct_field_mappings" do
+    # table is missing
```
paulcsmith: I think you can remove these comments since the `it` blocks does a good job of explaining it

spec/model_spec.cr
paulcsmith: I like that 👍

spec/model_spec.cr
```diff
@@ -97,4 +120,48 @@ describe LuckyRecord::Model do
...
+        updated_at: now,
+        missing: "missing"
+      expect_raises Exception, "The table users does not have a 'missing' column. Make sure you've added it to the migration." do
+        user.ensure_correct_field_mappings!
+      end
+    end
+
+    it "raises on nilable fields with required columns" do
+      now = Time.now
+      user = OptionalFieldOnRequiredColumn.new id: 1,
```
paulcsmith: Now that I'm reading this in the test I think it might be more clear if you add `ModelWith` to these models. Otherwise it looks like you're instantiating a Column object and not a model

spec/model_spec.cr
```diff
@@ -97,4 +120,48 @@ describe LuckyRecord::Model do
...
+        created_at: now,
+        updated_at: now,
+        name: "Mikias"
+      expect_raises Exception, "name is marked as nilable (name : String?), but the database column does not allow nils." do
```
paulcsmith: Can you wrap `name` with `'` single quotes so it stands out?

spec/model_spec.cr
```diff
@@ -97,4 +120,48 @@ describe LuckyRecord::Model do
...
+      end
+    end
+
+    it "raises on required fields with nilable columns" do
+      now = Time.now
+      user = RequiredFieldOnOptionalColumn.new id: 1,
+        created_at: now,
+        updated_at: now,
+        nickname: "Miki"
+      expect_raises Exception, "nickname is marked as required (nickname : String), but the database column it not." do
```
paulcsmith: `but the database column allows nils.`

spec/model_spec.cr
```diff
@@ -9,6 +9,29 @@ private class QueryMe < LuckyRecord::Model
   end
 end
 
+private class MissingColumn < LuckyRecord::Model
+  table users do
+    field missing : String
+  end
```
paulcsmith: What do you think about adding a feature for `Did you mean...` for similarly named columns? That might be helpful

src/lucky_record/model.cr
```diff
@@ -36,6 +39,7 @@ class LuckyRecord::Model
     setup_base_form_class({{table_name}})
     setup_table_name({{table_name}})
     setup_fields_method
+    setup_correct_field_mappings_method
```
paulcsmith: I think this is a large enough new feature that it would be good to extract a module for this. Something like this maybe:

```crystal
module LuckyRecord::SchemaEnforcer
  macro add_schema_enforcer_methods_for(fields, table_name)
    # generate all the methods that you were previously doing with `macro setup_correct_field_mappings_method`
  end
end
```

src/lucky_record/model.cr
```diff
@@ -103,4 +107,89 @@ class LuckyRecord::Model
...
+          { name: :{{field[:name]}}, nilable: {{ field[:nilable] }}, type: {{ field[:type] }} },
+        {% end %}
+      ]
+      ensure_fields_match_columns! fields
+    end
+  end
+
+  def ensure_existing_table!
+    table_names = LuckyRecord::Repo.tables_with_schema(excluding: "migrations")
+    return if table_names.includes?(@@table_name.to_s)
```
paulcsmith: Early returns can often cause misleading conditionals. Instead what if you extracted a class to handle this and use a regular if/else?

```crystal
# in the LuckyRecord::SchemaEnforcer module
class EnsureExistingTable
  def initialize(@table_name : Symbol)
    @table_names = LuckyRecord::Repo.tables_with_schema(excluding: "migrations")
  end

  def ensure_exists!
    if table_missing?
      # raise the error message
    end
  end

  private def table_missing?
    !@table_names.includes?(@table_name)
  end
end
```

I hope that makes sense :) LMK if I wasn't clear and I can try to clarify more

src/lucky_record/model.cr
```diff
@@ -103,4 +107,89 @@ class LuckyRecord::Model
...
+      end
+    end
+
+    if missing_columns.empty? && optional_field_errors.empty? &&required_field_errors.empty?
+      return
+    end
+
+    message = missing_columns + optional_field_errors + required_field_errors
+
+    raise message.join("\n\n")
```
paulcsmith: I think this could also be extracted to a class in `LuckyRecord::SchemaEnforcer` called `EnsureMatchingColumns`

Then you could extract some private methods for some of these to keep the class super easy to read. LMK if you want some help with this.

src/lucky_record/model.cr
paulcsmith: I think @hanneskaeufler is right. The second option should be `Or, change the column in a migration to be required: make_required: ...`. Since the column is already optional if it hits this case

src/lucky_record/repo.cr
```diff
@@ -17,41 +17,60 @@ class LuckyRecord::Repo
...
+    WHERE table_schema = 'public'
+      AND table_name = '#{table_name}'
+    SQL
+
+    run { |db| db.query_all statement, as: TableColumn }
+  end
 
-    def select_rows(statement)
-      rows = [] of String
+  class TableColumn
```
paulcsmith: 👍 I like this a lot. Nice abstraction

spec/model_spec.cr
mikeeus: ah yes I forgot to remove these

spec/model_spec.cr
mikeeus: Oh yeah that could be helpful

spec/model_spec.cr
mikeeus: I tried adding color, but it broke the test by surrounding the colorized text with `\e[33m` and `\e0m`.
```
Expected Exception with "The table 'uusers' was not found. \e[33mDid you mean users?\e[0m", got #<Exception: The table 'uusers' was not found. Did you mean users?
```
Do you know how to handle this?

src/lucky_record/model.cr
mikeeus: Oooooh, I completely missed your point @hanneskaeufler . I thought you meant that the message didn't need both `Mark 'field_name' as nilable...` and `Or, change the column...`. :man_facepalming:

mikeeus:
Thanks, this one was a lot of fun! I learned how to use `Levenshtein` and got good practice on testing and code organization. Do you have another issue you'd like me to tackle next?

Btw, good job on adding `crystal tool format` to the CI. I always forget to run it, but I'm sure it'll develop into a habit now.

paulcsmith:
Oh good. I'm glad you enjoyed it and got to learn some new stuff!

I also forget to run it and check it sometimes. I'm glad you found it helpful. I pinged you on a potential next issue :)

PR #164, opened 2017-12-17
Add first? and last? query methods that use existing implementation

### LINKED ISSUE #152: first?/last? Should limit to 1 record
Right now it loads all records and then gets the first one. It should work more like the existing “first method, but use “first?” So it returns nil if nothing was found:

```crystal
def first
  first || raise RecordNotFoundError...
end

def first?
  # Basically the existing `first` method without raising
  query.limit(1)
  exec_query.first?
end
```

mikeeus:
Closes #152 

I used the existing `first?` and `last?` methods and rescued `RecordNotFoundError` to avoid duplication.

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -30,6 +30,21 @@ describe LuckyRecord::Query do
     end
   end
 
+  describe "#first?" do
+    it "gets the first row from the database" do
+      UserBox.new.name("First").save
+      UserBox.new.name("Last").save
+
+      user = UserQuery.new.first?
+      user.should_not be_nil
+      user && user.name.should eq "First"
+    end
+
+    it "returns nil if no record found" do
+      UserQuery.new.first?.should be_nil
+    end
+  end
+
   describe "#last" do
     it "gets the last row from the database" do
       UserBox.new.name("First").save
@@ -50,6 +65,21 @@ describe LuckyRecord::Query do
       expect_raises(LuckyRecord::RecordNotFoundError) do
         UserQuery.new.last
       end
+    end
+  end
+
+  describe "#last?" do
+    it "gets the last row from the database" do
+      UserBox.new.name("First").save
+      UserBox.new.name("Last").save
+
+      last = UserQuery.new.last?
+      last.should_not be_nil
+      last && last.name.should eq "Last"
+    end
+
+    it "returns nil if last record is not found" do
+      UserQuery.new.last?.should be_nil
     end
   end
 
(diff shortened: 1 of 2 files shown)


spec/query_spec.cr
```diff
@@ -30,6 +30,21 @@ describe LuckyRecord::Query do
...
   end
 
+  describe "#first?" do
+    it "gets the first row from the database" do
+      UserBox.new.name("First").save
+      UserBox.new.name("Last").save
+
+      user = UserQuery.new.first?
+      user.should_not be_nil
+      user && user.name.should eq "First"
```
paulcsmith: You could also remove `user &&` and do `user.not_nil!.name.should eq "First"`. I'm fine with either, but wanted to let you know about the other option in case you prefer it

spec/query_spec.cr
```diff
@@ -30,6 +30,21 @@ describe LuckyRecord::Query do
...
+    end
+
+    it "returns nil if no record found" do
+      UserQuery.new.first?.should be_nil
```
paulcsmith: 👍

src/lucky_record/queryable.cr
```diff
@@ -50,11 +50,23 @@ module LuckyRecord::Queryable(T)
     exec_query.first? || raise RecordNotFoundError.new(model: @@table_name, query: :first)
   end
 
+  def first?
+    first
+  rescue RecordNotFoundError
```
paulcsmith: Raising and rescuing is a fairly expensive operation, so it's better to avoid it if possible. So instead I think this would be more performant if it looked a bit more like the issue, where `first?` and `last?` did something like this:

```crystal
# Basically the existing `first` method, but it doesn't raise
def first?
  query.limit(1)
  exec_query.first?
end	

def first
  first? || raise RecordNotFoundError.new(model: @@table_name, id: id.to_s)
end	      
```

This way we don't raise/rescue unnecessarily

mikeeus: Okay I wasn't sure about the performance penalty, this clears it up thanks

spec/query_spec.cr
mikeeus: oh cool I didn't realize I could chain `not_nil!`

src/lucky_record/queryable.cr
mikeeus: I guess I didn't read the issue properly because I completely missed the suggested implementation

paulcsmith:
Lookin good!

PR #165, opened 2017-12-17
Raise helpful error when count is used with offset or limit

### LINKED ISSUE #133: Show helpful error message if trying to count with limit/offset
For now, raise helpful error if trying to count with limit/offset

Later on add subquerying so they can be used. See comment:

https://github.com/luckyframework/lucky_record/issues/129#issuecomment-348685156

Once this is done, reopen https://github.com/luckyframework/lucky_record/issues/133

Message should be something like:

```
Counting with limit or offset is not supported yet. Try calling `results.size` to count in Crystal instead.
```

mikeeus:
Closes #133

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -158,12 +158,11 @@ describe LuckyRecord::Query do
     end
 
     it "raises when used with offset or limit" do
-      msg = "Counting with limit or offset is not supported yet. Try calling `results.size` to count in Crystal instead."
-      expect_raises Exception, msg do
+      expect_raises(LuckyRecord::UnsupportedQueryError) do
         UserQuery.new.limit(1).count
       end
 
-      expect_raises Exception, msg do
+      expect_raises(LuckyRecord::UnsupportedQueryError) do
         UserQuery.new.offset(1).count
       end
     end
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -156,6 +156,17 @@ describe LuckyRecord::Query do
 
       query.count.should eq 1
     end
+
+    it "raises when used with offset or limit" do
+      msg = "Counting with limit or offset is not supported yet. Try calling `results.size` to count in Crystal instead."
+      expect_raises Exception, msg do
+        UserQuery.new.limit(1).count
+      end
+
+      expect_raises Exception, msg do
+        UserQuery.new.offset(1).count
+      end
+    end
   end
 
   describe "#not with an argument" do
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -158,11 +158,11 @@ describe LuckyRecord::Query do
     end
 
     it "raises when used with offset or limit" do
-      expect_raises(LuckyRecord::UnsupportedQueryError) do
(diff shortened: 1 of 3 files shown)


mikeeus:
@hanneskaeufler I don't know, do you think it would be useful down the line?

hanneskaeufler:
Not particularly :) But general benefits of using named exceptions apply I guess. Easier to catch, and in the tests we dont have to be as strict with the message. `expect_raises Lucky::UnsupportedSql do` without asserting the message is more robust against tiny changes in the message, which I like.

mikeeus:
I get that. In that case I think `Exception` types for #163 would be really useful, those messages are long and testing for them is messy.

It's late for me but I'll make an `Exception` class for this tomorrow

spec/query_spec.cr
```diff
@@ -156,6 +156,17 @@ describe LuckyRecord::Query do
 
       query.count.should eq 1
     end
+
+    it "raises when used with offset or limit" do
+      msg = "Counting with limit or offset is not supported yet. Try calling `results.size` to count in Crystal instead."
+      expect_raises Exception, msg do
```
paulcsmith: I'm fine with the `UnsupportedSQLError` or this, but either way you can make these expectations simpler by using a subset of the message: `expect_raises Exception, "not support yet"`

This works because the `expect_raises` helper uses String#includes? instead of equality: https://github.com/crystal-lang/crystal/blob/204bfd0555921f3aadbda289993cca5323ebaf95/src/spec/expectations.cr#L272

This makes the tests a lot less brittle. I hope that little tidbit is helpful for you to know :)

src/lucky_record/query_builder.cr
```diff
@@ -86,6 +86,9 @@ class LuckyRecord::QueryBuilder
   end
 
   def count
+    if @limit || @offset
+      raise "Counting with limit or offset is not supported yet. Try calling `results.size` to count in Crystal instead."
```
paulcsmith: After reading this a bit more, I think `results.size` might not be that clear. What do you think about `Try calling 'results.size' on your query to count using Crystal instead of the database.` Thoughts?

mikeeus: I'm all for more clarity :+1:

spec/query_spec.cr
mikeeus: That's definitely a useful feature and I try to use it as much as possible. I guess it's enough for most cases, and the main point of using an `Exception` class is to be able to rescue it explicitly, which is good to know

src/lucky_record/errors.cr
```diff
@@ -25,4 +25,11 @@ module LuckyRecord
       super "Invalid #{form_name}. Could not save #{@form_object}"
     end
   end
+
+  # Raised when an unimplemented or deprecated query is made.
+  class UnsupportedQueryError < LuckyRecordError
```
paulcsmith: I like this description and error. I think it is generic, but the message is specific. So maybe it should accept a message so it can be used for other unsupported queries in the future:

```crystal
def initialize(message : String)
  raise message
end
```

PR #166, opened 2017-12-17
Run Form#insert and Form#update in a transaction

### LINKED ISSUE #50: Save inside of a transaction
https://github.com/crystal-lang/crystal-db/pull/27

mikeeus:
Closes #50

I just ran the insert and update methods in a transaction as is. I'm not sure how to test this but with some direction I can get it done.

### DIFF
--- a/spec/form_spec.cr
+++ b/spec/form_spec.cr
@@ -15,6 +15,16 @@ end
 private class TaskForm < Task::BaseForm
 end
 
+private class UserCallbackForm < User::BaseForm
+  allow :id, :name, :age, :joined_at
+
+  def before_save
+    first = UserQuery.new.first
+    UserForm.new(first, name: "changed").save! # This doesn't use the same tx connection
+    id.value = first.id
+  end
+end
+
 describe "LuckyRecord::Form" do
   it "generates the correct form_name" do
     LimitedUserForm.new.form_name.should eq "limited_user"
@@ -224,6 +234,21 @@ describe "LuckyRecord::Form" do
           record.should be_nil
         end
       end
+    end
+  end
+
+  describe "saves in a transaction" do
+    it "runs in a transaction" do
+      params = {"name" => "New Name", "age" => "30", "joined_at" => now_as_string}
+      before = UserForm.new(params).save!
+
+      params = {"name" => "name", "age": "26", "joined_at": now_as_string}
+      expect_raises Exception, "" do
+        UserCallbackForm.new(params).save!
+      end
+
+      pp before
+      UserQuery.new.first.name.should eq "New Name"
     end
   end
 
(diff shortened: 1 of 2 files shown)


mikeeus:
I have a better understanding of the problem now, and I wrote a test that should cover it, but implementing is going to be more complicated.

spec/form_spec.cr
```diff
@@ -15,6 +15,16 @@ end
 private class TaskForm < Task::BaseForm
 end
 
+private class UserCallbackForm < User::BaseForm
+  allow :id, :name, :age, :joined_at
+
+  def before_save
+    first = UserQuery.new.first
+    UserForm.new(first, name: "changed").save! # This doesn't use the same tx connection
```
mikeeus: In order to trigger the transaction rollback, the nested `UserForm` needs to access the `UserCallbackForm`s transaction.

One possibility may be to store the transaction in the `User::BaseForm` as a class method, but then it's restricted to classes which isn't useful.

I imagine ideal way would be to get the context of `UserForm.new(...).save!`, meaning the `before_save` method then getting the `UserCallbackForm` from there, but I don't even know if that's possible.

paulcsmith: I believe you can also do this:

https://github.com/crystal-lang/crystal-db/blob/28b17b7dba820d7e4810493aaf0a99e368849ec6/spec/transaction_spec.cr#L68

I'm not entirely sure if it works with nested transactions, but I think it is worth a shot! If it works that would be incredible. If not, LMK and we'll try to think of something. My guess would be adding a `transaction` property on forms that is `nil` by default, but if it is set, it uses it instead of creating a new transaction

mikeeus: So the idea is that raising `DB::Rollback.new` will bubble up to the next transaction? I'll try it out and see if I can make it work

paulcsmith: Yeah exactly. I'm fairly sure it should bubble *fingers crossed*

mikeeus: I'm still getting the same error. `UserForm` is creating its own transaction and committing it before the execution comes back to `before_save`. 

From a quick glance at the code, it looks like Rails handles transactions by tracking them in `ActiveRecord::Base`. We can do something similar in BaseForm, but what happens in a multi threaded situation where a user spawns fibers that might interact with the db. Wouldn't the fibers all have access to the same `BaseForm.transaction`? (I'm still new to concurrency with fibers)

paulcsmith:
I'm not 100% sure how to handle this so I'm going to branch of your branch and try some stuff out and hopefully figure some stuff out

paulcsmith:
Closing this in favor of https://github.com/luckyframework/lucky_record/pull/185

Thanks for starting on this @mikeeus. It was a tricky one!

PR #171, opened 2017-12-21
Allow raw where queries

### LINKED ISSUE #56: Add way to use raw WHERE query
So that if there is something not implemented by LuckyRecord, user's can still do what they need to do when joining, using OR, etc.

Something like:

```crystal
UserQuery.new.raw_where("users.name = #{PG::EscapeHelper.escape_literal("Paul")}") 
```

This would likely be done by adding a `@raw_where = [] of String` to `QueryBuilder` that gets joined to the regular where clauses. Then adding a `raw_where` method to `Queryable`

### Note

It will be helpful to add something to the guides that explains this AND explains how to generate a 100% custom query and parse with Lucky::Repo and DB.mapping. That's something that shouldn't require any code changes in LuckyRecord though.

mikeeus:
Closes #56 

Here's my go at it. I added an overload to `Queryable#where(statement : String)` because I thought it would be cleaner to match Rails' syntax rather than have a separate `raw_query` method.

I also raised an error when the number of bind variables doesn't match the number of arguments given. The error message is exactly the same as you find in Rails.


edwardloveall:
Nice work! This is a really straight forward API. You don't even have to think about which one to use!

BTW, I think your merge went funky.

src/lucky_record/queryable.cr
```diff
@@ -20,11 +20,16 @@ module LuckyRecord::Queryable(T)
...
   end
 
-  def where(column, value)
+  def where(column : Symbol, value)
     query.where(LuckyRecord::Where::Equal.new(column, value.to_s))
     self
   end
 
+  def where(statement : String, *args)
+    query.raw_where(LuckyRecord::Where::Raw.new(statement, *args))
```
paulcsmith: Maybe `bind_vars` instead of `args`?

src/lucky_record/where.cr
```diff
@@ -140,4 +140,37 @@ module LuckyRecord::Where
...
+    end
+
+    private def ensure_enough_bind_variables_for!(*args)
+      bind_vars = @statement.chars.select(&.== '?')
+      if bind_vars.size != args.size
+        raise "wrong number of bind variables (#{args.size} for #{bind_vars.size}) in #{@statement}"
+      end
+    end
+
+    private def build_clause(*args)
```
paulcsmith: And `bind_vars` here too?

src/lucky_record/where.cr
```diff
@@ -140,4 +140,37 @@ module LuckyRecord::Where
...
+    end
+
+    def to_sql
+      @clause
+    end
+
+    private def ensure_enough_bind_variables_for!(*args)
+      bind_vars = @statement.chars.select(&.== '?')
+      if bind_vars.size != args.size
+        raise "wrong number of bind variables (#{args.size} for #{bind_vars.size}) in #{@statement}"
```
paulcsmith: Nice! 👍

src/lucky_record/where.cr
```diff
@@ -140,4 +140,37 @@ module LuckyRecord::Where
...
   end
+
+  class Raw
+    @clause : String
+    getter :statement
+
+    def initialize(@statement : String, *args)
+      ensure_enough_bind_variables_for! *args
+      @clause = @statement
+      build_clause *args
```
paulcsmith: Instead of setting @clause and then calling build_clause, it would probably be simpler to do something like this:

```crystal
@clause = build_clause(statement, bind_vars)
```

That way you don't need to set the `@statement` variable or set up a getter for it. You can make it a local variable instead

mikeeus: nice, that's much cleaner!

spec/query_spec.cr
```diff
@@ -126,6 +126,20 @@ describe LuckyRecord::Query do
     end
   end
 
+  describe "#raw_where" do
```
paulcsmith: Oops, I just noticed this should be `#where`

spec/query_spec.cr
```diff
@@ -126,6 +126,20 @@ describe LuckyRecord::Query do
...
+    it "accepts multiple arguments and chains with #where and itself" do
+      query = UserQuery.new.where("first_name = ? AND last_name = ?", "Mikias", "Abera").where(:age, 26).query
+
+      query.statement.should eq "SELECT #{User::COLUMNS} FROM users WHERE age = $1 AND first_name = 'Mikias' AND last_name = 'Abera'"
```
paulcsmith: And could you also call `results` on the query like some of the other tests, just to make sure it runs successfully?

mikeeus:
I ran into an annoying bug:
```
could not determine data type of parameter $1
```
It was caused by calling `query.args` followed by `query.results` because both call the `wheres_sql` which in turn calls `next_prepared_statement_placeholder` twice. This increments the placeholder to `$2` while there's still only one variable.

I solved it by setting the sql once with `@wheres_sql ||= joined_wheres_clauses`.

src/lucky_record/query_builder.cr
```diff
@@ -148,6 +149,10 @@ class LuckyRecord::QueryBuilder
   end
 
   private def wheres_sql
+    @wheres_sql ||= joined_wheres_queries
```
paulcsmith: Nice catch!

paulcsmith:
Closed by https://github.com/luckyframework/lucky_record/commit/ba7d5f6c5ce8c3a3c3fd376d2c16aad681ba244d

PR #172, opened 2017-12-21
Add has one association

### LINKED ISSUE #168: Add has_one
Similar to has_many, but calls `first` or `first?` depending on if it is nilable or not

```crystal
class User < BaseModel
  table :users do
    has_one sign_in_credential : SignInCredential
  end
end
```

mikeeus:
Closes #168 

I added a `has_one` association macro which works when I tested it manually by adding `has_one sign_in_credential : SignInCredential` and `has_one sign_in_credential : SignInCredential?` to the `User` class. But I'm having trouble writing specs for it.

I tried creating a class `CredentialedUser` that inherits from `User` so I could use the `UserBox`:
```crystal
ss CredentialedUser < User
  table users do
    has_one sign_in_credential : SignInCredential
  end
```
this seemed to work, but the method that the `has_one` macro generates gets the foreign_key from the class name:
```crystal
def {{ assoc_name.id }}
  {{ model }}::BaseQuery.new.{{ @type.name.underscore }}_id(id).first{% if nilable %}?{% end %}
end
```
so the `user.sign_in_credential` query gets messed up:
```bash
   9. 
  10.     def sign_in_credential
> 11.       SignInCredential::BaseQuery.new.credentialed_user_id(id).first
  12.     end
```
How should I be testing a case like this?


mikeeus:
Also, I don't know why my pull requests are referencing commits from 5 days ago.

spec/associations_spec.cr
```diff
@@ -1,5 +1,11 @@
 require "./spec_helper"
 
+# class CredentialedUser < User
```
paulcsmith: Instead of inheriting from user, you can inherit from the model and only add a `has_one`. Not all fields need to be defined for every model, so that should work

db/migrations/20170127143150_create_sign_in_credentials.cr
```diff
@@ -0,0 +1,13 @@
+class CreateSignInCredentials::V20170127143150 < LuckyMigrator::Migration::V1
+  def migrate
+    create :sign_in_credentials do
+      add user_id : Int32
+      add email : String
+      add encrypted_password : String
```
paulcsmith: Usually when writing test it is best to only add what is necessary to verify everything works. So in this case you could remove email and encrypted_password

paulcsmith:
Oh and for those extra pull requests you can do: git rebase -i master

Then delete the lines for the old commits and force push to your branch. That should do the trick

mikeeus:
I ran `git rebase -i master` and it opened an empty file, so I just exited without saving..

paulcsmith:
@mikeeus You may need to checkout master and `git pull origin master` to make sure master is up to date. If that doesn't work though it should be fine. I can squash and merge it once the code is ready

mikeeus:
I tried that and same result so I'll just leave it as is then

spec/associations_spec.cr
mikeeus: Sorry but I don't understand. Do you mean inherit from `LuckyRecord::Model`? Because that gives me the same issue as before: `SignInCredential::BaseQuery.new.credentialed_user_id(id).first`

paulcsmith: Ohh, I see what you mean now. I think the problem then, is that we need to add an option for setting the foreign_key: `has_one sign_in_credential : SignInCredential, foreign_key: :user_id`

But for now, I think the solution would be to create a different table (not users) and a different model. Maybe `Admin` and an `:admins` table. 

Does that make sense?

mikeeus: Absolutely, I actually started adding a foreign_key option earlier but I thought I was complicating it. Good to know I was on the right track.

So now I'll add an Admin table where lets say credentials are required, and make the user credentials optional. Thus testing both?

paulcsmith: Yeah I think that makes the most sense 👍

mikeeus: K I tried this too, but the method still looks for: `SignInCredential::BaseQuery.new.admin_id(id).first` so I either have to add both an `admin_id` and `user_id` which is a bad idea.

So I just added the `foreign_key` option to the `has_one` macro which wasn't difficult since I already added the option for `belongs_to`.

spec/support/admin.cr
```diff
@@ -0,0 +1,13 @@
+require "./sign_in_credential"
+
+class Admin < LuckyRecord::Model
+  COLUMNS = "admins.id, admins.created_at, admins.updated_at, admins.name"
+
+  table admins do
+    field name : String
+    has_one sign_in_credential : SignInCredential, foreign_key: :user_id
```
paulcsmith: 👍

spec/support/boxes/admin_box.cr
```diff
@@ -0,0 +1,13 @@
...
+    name "Admin"
+  end
+
+  def build_model
+    Admin.new(
+      id: 123,
+      created_at: Time.now,
+      updated_at: Time.now
+    )
+  end
```
paulcsmith: I don't think you need this method

src/lucky_record/associations.cr
```diff
@@ -10,6 +10,31 @@ module LuckyRecord::Associations
...
+      query = {{ model }}::BaseQuery.new
+      {% if foreign_key %}
+        query.{{ foreign_key.id }}(id)
+      {% else %}
+        query.{{ @type.name.underscore }}_id(id)
+      {% end %}
+
+      query.first{% if nilable %}?{% end %}
+    end
+  end
```
paulcsmith: I love the new foreign_key option 👍

paulcsmith:
I think this just needs a rebase and formatting :D Thanks for doing this.

paulcsmith:
To get the rebase to work you may need to do this:

1. Checkout master
1. `git pull --rebase git@github.com:luckyframework/lucky_record.git master`. This will get the latest changes from lucky_record
1. checkout your feature branch
1. git rebase -i master

Ping me in Gitter if you run into trouble. Sometimes this is a bit tricky.

paulcsmith:
This may also be helpful: https://stackoverflow.com/questions/7244321/how-do-i-update-a-github-forked-repository

The problem is likely that your fork is out of date

mikeeus:
Thanks `git rebase -i master` works now! I tried removing the old commits and squashing the new ones. I think I did it right :crossed_fingers:

spec/associations_spec.cr
```diff
@@ -1,5 +1,11 @@
...
+#   table users do
+#     has_one sign_in_credential : SignInCredential
+#   end
+# end
```
paulcsmith: Do we need this model anymore or can it be removed?

mikeeus: I forgot about that one, its removed now

paulcsmith:
Nice! 🎉

PR #124, opened 2017-11-30
Pass Int32 to db as integer

mikeeus:
Closes #122.

PR #176, opened 2017-12-22
Add `foreign_key` argument to `has_many` association

### LINKED ISSUE #173: Add foreign_key option to belongs_to, has_one, has_many

mikeeus:
Closes #173

Looks like the rebase didn't solve the issue.

I tried changing the base branch to something else and changing it back based on a [stackoverflow answer](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch), but it didn't do the trick.

If we can finish up #124, then I can just refork the repo.


src/lucky_record/associations.cr
```diff
@@ -1,12 +1,19 @@
...
-  macro has_many(type_declaration)
+  macro has_many(type_declaration, foreign_key = nil)
     {% assoc_name = type_declaration.var %}
 
-    association table_name: :{{ assoc_name }}
+    association table_name: :{{ assoc_name }}, foreign_key: {{ foreign_key }}
 
     {% model = type_declaration.type %}
     def {{ assoc_name.id }}
-      {{ model }}::BaseQuery.new.{{ @type.name.underscore }}_id(id)
```
mikeeus: This was returning `BaseQuery` rather than the model, but the tests weren't catching it. I changed it to return `results` which I think is the expected behavior

mikeeus: Or actually is it supposed to return the `Query` so that it can be chainable. Like: `manager.employees.company_id(10)`?

paulcsmith:
@mikeeus I made a bunch of changes to this file when I did preloading so this may be a fairly complex rebase. LMK if you need some help with it

mikeeus:
The rebase looked straight forward considering you added support for the `foreign_key` in the `lazy_load_{{assoc_name}}` method. But now I'm getting the following really weird error when running the specs:

```
Error in src/lucky_record/queryable.cr:81: expanding macro

  getter preloads = [] of Array(T) -> Nil
  ^

in macro 'getter' expanded macro: macro_140501379124848:113, line 4:

   1.       
   2.         
   3.           
>  4.             @preloads = [] of (Array(T) -> Nil)
   5. 
   6.             def preloads
   7.               @preloads
   8.             end
   9.           
  10.         
  11.       
  12.     

instance variable '@preloads' of User::BaseQuery must be Array(Proc(Array(User), Nil)), not Array(Proc(Array(User), Nil))
```
It says `Array(Proc(Array(User), Nil))` != `Array(Proc(Array(User), Nil))` which is obviously not true. Do you have any idea why this would happen?

paulcsmith:
@mikeeus That is very strange. The error must be something else, but the compiler is incorrectly blaming it on the Proc. I did a few tests and couldn't figure it out right away. I'll have to take a closer look later this week

Issue #177, opened 2017-12-23
Setting a nilable field to nil through a form errors


hanneskaeufler:
Say you have a model with a nilable date field

```crystal
class Post < BaseModel
  table :posts do
    field published_at : Time?
  end
end
```

And a form that allows setting the field to nil with a checkbox

```crystal
class PostForm < Post::BaseForm
  allow_virtual published : Bool

  def prepare
    published_at.value = published.value ? Time.now : nil
  end
end
```

Submitting the form with the checkbox unchecked (so setting the field to `nil`) will return the following error:

`ERROR: invalid input syntax for type timestamp with time zone: "" (PQ::PQError)`

hanneskaeufler:
See #178 for my proposed fix, however I hope there's a better way other then to nil-check.

paulcsmith:
@hanneskaeufler I made a PR similar to yours here: https://github.com/luckyframework/lucky_record/pull/180 and merged it in. Thanks for pointing this out! It worked ok with nil string fields by setting them to an empty string, but that of course is not desired, and it would cause failures on anything other than string columns. 

LMK if this latest version change fixes it :)

hanneskaeufler:
Cool, thanks! Still haven't figured out how to pin lucky's transient dependencies to a specific version so will have to wait on a version bump in lucky_record and lucky to try this :)

PR #178, opened 2017-12-23
Updating/saving a nilable field to nil results in a postgres error.

hanneskaeufler:
This tries to reproduce /  fix the error in #177. The changes to the production code fixes the issue for me locally. However the changes aren't pretty, and the tests aren't finished either. I just wanted to get this out there in order to make the issue clearer.


paulcsmith:
Went with a similar approach in https://github.com/luckyframework/lucky_record/pull/180

Thanks for reporting and for the inspiration in this PR 👍

PR #185, opened 2018-01-03
Transaction support

### LINKED ISSUE #50: Save inside of a transaction

paulcsmith:
Closes #50

### DIFF
--- a/src/lucky_record/repo.cr
+++ b/src/lucky_record/repo.cr
@@ -1,5 +1,8 @@
 class LuckyRecord::Repo
+  alias FiberId = UInt64
+
   @@db : DB::Database? = nil
+  private class_getter transactions = {} of FiberId => DB::Transaction
 
   Habitat.create do
     setting url : String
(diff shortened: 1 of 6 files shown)


src/lucky_record/repo.cr
```diff
@@ -1,23 +1,57 @@
...
+    if current_transaction
+      yield
+    else
+      wrap_in_transaction do
+        yield
+      end
+    end
+  end
+
+  def self.wrap_in_transaction
```
paulcsmith: @mikeeus I saved the current transaction to a map with the current fiber id as the key. That gets it working with nested transactions, and avoids passing the transaction all the time.

Let me know if you have any questions about the implementation!

mikeeus: This is super cool! I knew the implementation would need to use `Fibers` but I wasn't sure how.

PR #194, opened 2018-01-09
Add foreign_key argument to has_many association

mikeeus:
I reworked this in a new branch and it seems that the problem occurs when inheriting from User. 
```crystal
class KeyHolder < User
  ...
end
```
So I don't think this bug is related to the associations specifically. I'll put the error here for reference:
```
Error in src/lucky_record/queryable.cr:81: expanding macro

  getter preloads = [] of Array(T) -> Nil
  ^

in macro 'getter' expanded macro: macro_140501379124848:113, line 4:

   1.       
   2.         
   3.           
>  4.             @preloads = [] of (Array(T) -> Nil)
   5. 
   6.             def preloads
   7.               @preloads
   8.             end
   9.           
  10.         
  11.       
  12.     

instance variable '@preloads' of User::BaseQuery must be Array(Proc(Array(User), Nil)), not Array(Proc(Array(User), Nil))
```


mikeeus:
If this is good to go, I'm going to close the other pull request and refork this repo to get rid of these commits.

src/lucky_record/associations.cr
```diff
@@ -3,16 +3,34 @@ module LuckyRecord::Associations
     LuckyRecord::Repo.settings.lazy_load_enabled
   end
 
-  macro has_many(type_declaration)
+  macro has_many(type_declaration, foreign_key = nil)
     {% assoc_name = type_declaration.var %}
 
     association table_name: :{{ assoc_name }}
```
paulcsmith: Does this also need to use a foreign_key option?

mikeeus: Do you mean custom foreign keys on has_many assocations? I don't have a use for them but I imagine this could be useful for someone's use case.

paulcsmith: Never mind. I was thinking about this wrong. Looks great

Issue #59, opened 2017-10-16
Query associations


paulcsmith:
Ideas for how it would look: https://gist.github.com/paulcsmith/0a511ab3db994e7b43ad48b8e3dba4a2

Blocked by: https://github.com/luckyframework/lucky_record/issues/128

paulcsmith:
Moving this to "Next" milestone since this can be done using the existing joins methods and #56 once that is done

PR #196, opened 2018-01-12
Add `has_many through` association

### LINKED ISSUE #61: Add has_many through option
A nice look at how Rails does it, some potential optimization’s we can look at, and what the generated SQL looks like: http://garbled.benhamill.com/2011/08/25/a-smarter-has_many-through/

Basically this option joins the intermediate table so you can include deeply nested associations:

```crystal
# Post model
has_many taggings : Taggings


# Taggings model
belongs_to tag : Tag
belongs_to post : Post

# Get the association
Post.new.first.tags # Returns tags for this post
```

> #135 may or may not be helpful to look at for ideas

mikeeus:
I added a through option that works for single level associations for ex:

```crystal
class Employee
    has_many :posts
    has_many :comments, through: :posts
end
```
But this doesn't work for anything deeper like:
```crystal
class Manager
    has_many :employees
    has_many :comments, through: :employees
end
```
Although this works, I feel like there could be a better solution taking advantage of `LuckyRecord::Model::ASSOCIATIONS` to track multiple levels of associations. I played around with it and couldn't get anything to work, so I'll wait for input before continuing.

Closes #61


mikeeus:
I tried working through this for a while now but I'm still not sure how to do it.

One issue I have is with the `belongs_to` association on the nested association. Right now, `belongs_to` calls `column` which we don't want on these associations because a column doesnt exist. For example:
```crystal
class Employee
    has_many posts : Post
    has_many comments : Comment, through: :posts
end

class Comment
    belongs_to post : Post
    belongs_to employee : Employee # through post
end
```
Now `CommentForm` expects an `employee_id` column, which there isn't.

paulcsmith:
@mikeeus Thanks for taking a crack at this. I'm guessing LuckyRecord will need https://github.com/luckyframework/lucky_record/issues/59 first. That way you can query by the association when joining. I'll take a crack at that and ping you when it's done!

paulcsmith:
Oops, I didn't mean to close this! #203 adds querying associations. So I think if you use `join` and the new methods for querying associations, you can get this to work. LMK if you need some help with it though. I'm happy to help

paulcsmith:
Thanks for your start on this @mikeeus. I know you're busy so I started work on this in another rbanch and opened: https://github.com/luckyframework/lucky_record/pull/206

mikeeus:
kk, thanks for understand Paul :)

PR #202, opened 2018-01-13
Add support for Int64 and Float64 column types

mikeeus:
I added support for both `Int64` and `Float64` because the process was similar.

Let me know if there's anything missing, I'd like to use this in a project right now. And once this is merged I can start working on using `Int64` for ids.


mikeeus:
Nice, it's satisfying to see this merged :D

PR #204, opened 2018-01-17
Add Queryable class methods for find, first and last

### LINKED ISSUE #97: Add class method shortcuts for find/first/last
So you can do `UserQuery.find(id)`, `UserQuery.first/last`

jdbann:
Closes #97

### DIFF
--- a/src/lucky_record/queryable.cr
+++ b/src/lucky_record/queryable.cr
@@ -10,11 +10,29 @@ module LuckyRecord::Queryable(T)
         queryable.query = query
       end
     end
-  end
 
-  macro included
     def self.all
       new
+    end
+
+    def self.find(id)
+      new.find(id)
+    end
+
+    def self.first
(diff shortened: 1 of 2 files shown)


src/lucky_record/queryable.cr
```diff
@@ -10,12 +10,30 @@ module LuckyRecord::Queryable(T)
         queryable.query = query
       end
     end
-  end
 
-  macro included
```
paulcsmith: Nice catch 👍

jdbann:
Can't see a way to mock without adding a new package, otherwise I'd far prefer to make sure that these methods just delegate to a new instance.

paulcsmith:
Yeah normally I'd say these tests are a bit brittle and a mock that tests delegation would be better, but find/last/first/etc are *extremely* unlikely to change so I think this should be totally fine

Issue #112, opened 2017-11-19
More validations


Xosmond:
### Currently we only have 
- validate_required
- validate_acceptance_of
- validate_confirmation_of ** Need documentation on the guides
- validate_inclusion_of       ** Need documentation on the guides

### We need:
- [x]  validate_uniqueness_of
- [ ] validate_format_of
- [x] validate_size_of
- validate_numericality_of (still debating best way to handle this)

Do we need an PR for each validation of a complete with all validations PR ?

paulcsmith:
Great suggestion. I think one PR for each would be the best way to handle this. I also think we don't need validate_numericality_of, at least not how it is in Rails. The reason is that each field is parsed by Lucky and if it can't be parsed to the given type it gets an "is invalid" message. So if you have an Int32, you basically get number validation. I think instead `validate_length_of` should be `validate_size_of` and would accept Strings or Int32 | Int64 | Float

```crystal
def validate_size_of(field : Field(String) | AllowedField(String), is expected_size : Int32)
  if field.value.try(&.size) != expected_size
    field.add_error "must be #{expect_size} characters long"
  end
end

def validate_size_of(field : Field(Int32 | Int64 | Float) | AllowedField(Int32 | Int64 | Float), is expected_size : Int32 | Int64 | Float)
  if field.value != expected_size
    field.add_error "must be #{expected_size}"
  end
end
```

validate_uniqueness_of would similarly to Rails where it queries the table. You should still use a unique index of course, but I like how Rails queries the database. When ORMs rely on the database constraint it means you will only get an invalid record when you actually try to insert the record. This leads to a poor user experience when your form has multiple errors, you fix them, submit it, then you get another uniqueness validation later. That's frustrating UX IMO.

Does that all make sense?

Xosmond:
## validate_numericality_of and validate_length_of
I understand your point in `validate_numericality_of` but I think there are other situations where we need that validation, for example:
- Can be used on validating fields that are Strings but they need to be numbers (Example: A Credit Card Number or A Passport Card Number)
- Can be used on validating fields that could be `String` or `Int` with options like:
  - only_integer ** This can't be done with `validate_size_of`
  - odd ** This can't be done with `validate_size_of`
  - even ** This can't be done with `validate_size_of`
  - greater_than
  - greater_than_or_equal_to
  - equal_to
  - less_than
  - less_than_or_equal_to

[...]

PR #205, opened 2018-01-17
Add size validation

jdbann:
One of two validations requested in #112

### DIFF
--- a/src/lucky_record/validations.cr
+++ b/src/lucky_record/validations.cr
@@ -26,4 +26,26 @@ module LuckyRecord::Validations
       field.add_error message
     end
   end
+
+  private def validate_size_of(field, *, is exact_size, message = "is invalid")
+    if field.value.to_s.size != exact_size
+      field.add_error message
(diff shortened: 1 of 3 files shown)


src/lucky_record/validations.cr
```diff
@@ -26,4 +26,26 @@ module LuckyRecord::Validations
       field.add_error message
     end
   end
+
+  private def validate_size_of(field, *, is exact_size, message = "is invalid")
```
paulcsmith: Can the `*` be removed? Maybe I'm missing something, but I'm not sure I understand it

jdbann: This was necessary to allow the subsequent `def validate_size_of` to overload. The `*` splat means all subsequent params need to be passed as named params.

To be explicit, I could add the splat to the second definition as well?

paulcsmith: Oh cool! I didn't realize you could do that.

PR #206, opened 2018-01-19
Has many through and Queryable#distinct

### LINKED ISSUE #61: Add has_many through option

paulcsmith:
Closes #61

### DIFF
--- a/src/lucky_record/associations.cr
+++ b/src/lucky_record/associations.cr
@@ -3,7 +3,7 @@ module LuckyRecord::Associations
     LuckyRecord::Repo.settings.lazy_load_enabled
   end
 
-  macro has_many(type_declaration, foreign_key = nil)
+  macro has_many(type_declaration, through = nil, foreign_key = nil)
     {% assoc_name = type_declaration.var %}
 
     association table_name: :{{ assoc_name }}, type: {{ type_declaration.type }}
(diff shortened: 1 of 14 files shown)


src/lucky_record/associations.cr
```diff
@@ -39,7 +53,26 @@ module LuckyRecord::Associations
...
+              end
+              .preload_{{ through.id }}
+              .distinct
+            {{ assoc_name }} = {} of Int32 => Array(Tag)
+            just_{{ assoc_name }}.each do |item|
+              item.{{ through.id }}.each do |item_through|
+                {{ assoc_name }}[item_through.{{ foreign_key }}] ||= Array({{ model }}).new
+                {{ assoc_name }}[item_through.{{ foreign_key }}] << item
+              end
+            end
```
paulcsmith: Consider extracting an object for this, though I'm not sure it would work because these are compile-time macros, so I don't know what you could pass to the class to make it work. I'll have to think about this some more

PR #212, opened 2018-02-07
Improve some error messages

### LINKED ISSUE #210: Better errors for wrong or missing allowed field on forms
I had a situation like this:

```crystal
class Shard < BaseModel
  table :shards do
    # other fields
    belongs_to category : Category?
  end
end

class Shards::New < BrowserAction
  action do
    render NewPage, shard_form: ShardForm.new
  end
end

class ShardForm < Shard::BaseForm
  allow category_id
  allow_virtual repo_name : String
  # other fields
end

class Shard::NewPage < MainLayout
  private def render(f : ShardForm)
    form_for Create do
      # other fields
      select_input f.category_id do
        options_for_select(f.category_id, categories_for_select)
      end

      submit "Add"
    end
  end
end
```

At various times, I had the `category_id` field in the ShardForm labeled incorrectly (`allow category_id : Int32?`) or missing. I got an error like this:

>**no overload matches 'Shards::NewPage#select_input' with type LuckyRecord::Field(Int32 | Nil)**
>Overloads are:
> \- Lucky::SelectHelpers#select_input(field : LuckyRecord::AllowedField, **html_options, &block)

It took me a while to figure out that `category_id` didn't need a type along with it. Perhaps a better error message would have led me in the right direction sooner.

paulcsmith:
Closes #210

### DIFF
--- a/src/lucky_record/form.cr
+++ b/src/lucky_record/form.cr
@@ -165,6 +165,27 @@ abstract class LuckyRecord::Form(T)
 
   macro allow(*field_names)
     {% for field_name in field_names %}
+      {% if field_name.is_a?(TypeDeclaration) %}
+        {% raise <<-ERROR
+          Must use a Symbol or a bare word in 'allow'. Instead, got: #{field_name}
+
+          Try this...
+
+            ▸ allow #{field_name.var}
+
+          ERROR %}
+      {% end %}
+      {% unless field_name.is_a?(SymbolLiteral) || field_name.is_a?(Call) %}
+        {% raise <<-ERROR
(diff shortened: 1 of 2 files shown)


src/lucky_record/form.cr
```diff
@@ -165,6 +165,27 @@ abstract class LuckyRecord::Form(T)
 
   macro allow(*field_names)
     {% for field_name in field_names %}
+      {% if field_name.is_a?(TypeDeclaration) %}
```
paulcsmith: If you do something like `allow name : String` it will tell you to do `allow name` instead :)

PR #243, opened 2018-06-17
Set updated_at during update

### LINKED ISSUE #241: updated_at only set during creation
I think it should be set every time the record is updated too, right?

hfjallemark:
Fixes #241

### DIFF
--- a/spec/form_spec.cr
+++ b/spec/form_spec.cr
@@ -262,6 +262,16 @@ describe "LuckyRecord::Form" do
       end
     end
 
+    it "allows overriding updated_at and created_at on create" do
+      user = UserBox.new
+        .created_at(Time.new(2018, 1, 1, 10, 20, 30))
+        .updated_at(Time.new(2018, 1, 1, 20, 30, 40))
+        .create
+
+      user.created_at.should eq Time.new(2018, 1, 1, 10, 20, 30)
+      user.updated_at.should eq Time.new(2018, 1, 1, 20, 30, 40)
+    end
+
     context "on success" do
       it "yields the form and the saved record" do
         params = {"joined_at" => now_as_string, "name" => "New Name", "age" => "30"}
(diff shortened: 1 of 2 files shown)


hfjallemark:
Makes sense! Will give it a try tomorrow.

hfjallemark:
I can't seem to get the `PostBox` to set `updated_at` properly: 

```crystal
post = PostBox.new.updated_at(1.day.ago).create
puts post.updated_at # Shows the current date and time, not yesterday's
```

Any ideas what I am doing wrong?

Here is the full test I currently have:

```crystal
it "updates updated_at" do
  post = PostBox.new.updated_at(1.day.ago).create
  params = {} of String => String
  PostForm.update post, with: params do |form, record|
    form.saved?.should be_true
    record.updated_at.should be > 1.second.ago
  end
end
```

paulcsmith:
You found another bug :)

The problem is here:

https://github.com/luckyframework/lucky_record/blob/4d2f58cc9857e63d5fc0f5ecbc44a46fc0ccef9d/src/lucky_record/form.cr#L322-L323

On save, it overrides created_at and updated_at. So your code is correct, but that methods need to use `||=` so that it doesn't override values set by the user.

Since this also affects `created_at` it might be good to have another spec like `it "allows overriding updated_at and created_at on create"`

hfjallemark:
Fixed bug and added spec, let me know if it's ready to merge @paulcsmith!

spec/form_spec.cr
```diff
@@ -262,6 +262,16 @@ describe "LuckyRecord::Form" do
...
 
+    it "allows overriding updated_at and created_at on create" do
+      user = UserBox.new
+        .created_at(Time.new(2018, 1, 1, 10, 20, 30))
+        .updated_at(Time.new(2018, 1, 1, 20, 30, 40))
+        .create
+
+      user.created_at.should eq Time.new(2018, 1, 1, 10, 20, 30)
+      user.updated_at.should eq Time.new(2018, 1, 1, 20, 30, 40)
+    end
```
paulcsmith: 👍

PR #245, opened 2018-06-19
First stab at callback docs

edwardloveall:
Since these methods are all so similar, I decided to just write class documentation. Maybe I could put some specifics on each method to specify when they are called? Perhaps also reiterating that the `after_*` methods take the saved record?

### DIFF
--- a/src/lucky_record/callbacks.cr
+++ b/src/lucky_record/callbacks.cr
@@ -1,3 +1,34 @@
+# Callbacks will trigger before and after the `create`, `save`, or `update` method is called on a form.
+#
+# To define a callback, you assign a method to one of the callback groups:
+#
+# ```crystal
+# class UserForm < User::BaseForm
+#   before_create set_token
+#   after_create send_welcome_email
+#
+#   def set_token
+#     token.value = # generate token
+#   end
+#
+#   def send_welcome_email(user : User)
+#     # send the user a welcome email
+#   end
+# end
+# ```
+#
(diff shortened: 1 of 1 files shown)


src/lucky_record/callbacks.cr
```diff
@@ -1,3 +1,34 @@
+# Callbacks will trigger before and after the `create`, `save`, or `update` method is called on a form.
+#
+# To define a callback, you assign a method to one of the callback groups:
+#
+# ```crystal
+# class UserForm < User::BaseForm
+#   before_create set_token
```
paulcsmith: The problem with `set_token` is that something like that should almost always happen in `prepare`

How about something like `before_create log_user_creation`. Probably not the best example, but that's probably ok since *before* should rarely be used.

It might also be worth mentioning that *before* is rarely used and that validations and setting values should usually happen in `prepare`

src/lucky_record/callbacks.cr
```diff
@@ -1,3 +1,34 @@
...
+# end
+# ```
+#
+# You'll notice that the example after method takes a `User`. All `after_*` callbacks take the newly saved record. In addition validation happens before all callbacks. If you need to set values before validation, override the `prepare` method on the form.
+#
+# When a triggering method is called (`save`, `create`, or `update`) the callbacks in each group execute in the order they were defined.
+#
+# The callback groups are called in the following order. The callbacks are highlighted in **bold**:
+#
+# * _validation_
```
paulcsmith: Might be worth mentioning this happens in `prepare`

edwardloveall:
Waiting on https://github.com/luckyframework/lucky_record/issues/261

paulcsmith:
I think we'll go with changing how callbacks work. Something like is mentioned in #261.

Issue #246, opened 2018-06-20
Using alternate primary key types (such as UUID)


jwoertink:
Sometimes you may want to use UUID or a custom string for your `ID` column as your primary key. Right now if you try to override this, lucky blows up because of [this](https://github.com/luckyframework/lucky_record/blob/master/src/lucky_record/model.cr#L13). Having a way to customize this would be pretty sweet.

In my case, I have an app that has to run in parallel with another app. The other app uses integer IDs. When a record comes in like `/sites/1`, I know this is data from the other app. When it comes in like `/sites/abc`, I know it's from this app. Using a UUID as an ID can also help to avoid people looking in to how many records you may have, or doing a hack like trying to access a record by incrementing the id in a route.

sfcgeorge:
Rails 5.1 allows configuring primary key type. By default it's still a number but [now supports UUID](https://lab.io/articles/2017/04/13/uuids-rails-5-1/). Apart from the metadata leaking point above, UUIDs also potentially allow sharding and whatnot.

Rails also changed the default numerical primary key type from SQL "integer" to "bigint". Because if you have something like a log table, it is possible to exhaust all integers then all of a sudden everything breaks (a company experienced this, which prompted the change). So the default should at least be changed from Int32 to Int64.

paulcsmith:
👍 LuckyRecord should definitely support UUID and setting a default of UUID would be great too.

I'm imaging that for one-off tables it would look like this:

```
table :users, primary_key: :uuid do
end

# Or if you set it to default to :uuid
table :users, primary_key: :bigint do
end
```

I think making it a default is a bit trickier so should be a separate issue, but making it configurable in the `table` macro should be fairly straightforward.

PR #254, opened 2018-06-25
UUID support as primary key

hanneskaeufler:
I wanted to get the ball rolling regarding UUIDS because ... it's interesting and I want to learn something by it (does that count as an excuse?). #246 got me excited. Feedback is appreciated at any time.

### How would this work from a user's perspective?

Define the model with `primary_key_type: :uuid` (this is how rails calls it I think)
```crystal
class LineItem < LuckyRecord::Model
  table :line_items, primary_key_type: :uuid do
    column name : String
  end
end
```

### Where are the ids generated?

From what I can see, we could either generate the uuids in crystal code (done right now, as it was easier), or through postgresql extensions.
I have no real grasp on availability of postgres extensions, so cannot comment on how realistic it would be to require that.

### Todos

- [x] Migrations https://github.com/luckyframework/lucky_migrator/pull/112
- [x] Reuse or add similar enum for primary id type
- [x] Spec for failed cast
- [x] Using actual crystal uuids or a wrapper thereof (?)


spec/support/boxes/line_item_box.cr
```diff
@@ -0,0 +1,8 @@
+require "uuid"
```
paulcsmith: I'd consider requiring this in `src/lucky_record.cr` so it is always available and you don't need to manually require it in other files

spec/spec_helper.cr
```diff
@@ -13,6 +13,6 @@ LuckyRecord::Repo.configure do
   end
 end
 
-Spec.after_each do
+Spec.before_each do
```
paulcsmith: Nice change!

src/lucky_record/charms/uuid_extensions.cr
```diff
@@ -0,0 +1,52 @@
...
+    alias ColumnType = UUID
+    include LuckyRecord::Type
+
+    def parse(value : UUID)
+      SuccessfulCast(UUID).new(value)
+    end
+
+    # form form #set_id_from_param
+    def parse(value : String)
+      SuccessfulCast(UUID).new(UUID.new(value))
```
paulcsmith: I think a spec that gives it a bad UUID would be good. Then this should catch any errors raised if the UUID is incorrect and return a `FailedCast`

spec/support/boxes/line_item_box.cr
```diff
@@ -0,0 +1,8 @@
...
+
+class LineItemBox < BaseBox
+  def initialize
+    id  UUID.random.to_s
```
paulcsmith: I think this should be removed since ideally either LuckyRecord, or the database should generate it. I don't think the developer should have to generate it themselves

src/lucky_record/model.cr
```diff
@@ -32,7 +40,7 @@ class LuckyRecord::Model
     end
   end
 
-  macro setup(table_name)
+  macro setup(table_name, primary_key_type)
```
paulcsmith: Looks good!

src/lucky_record/model.cr
```diff
@@ -1,16 +1,25 @@
...
   include LuckyRecord::SchemaEnforcer
 
   macro inherited
     FIELDS = [] of {name: Symbol, type: Object, nilable: Bool, autogenerated: Bool}
     ASSOCIATIONS = [] of {name: Symbol, foreign_key: Symbol, type: Object}
+  end
+
+  macro setup_autogenerated_columns(primary_key_type)
+    {% if primary_key_type == :uuid %}
+      column id : UUID, autogenerated: true
```
hanneskaeufler: @paulcsmith  The `UUID` here is what I want to make work. In contrast to `String` previously.

paulcsmith: Gotcha, and where is it still treating id as an `Int32`?

hanneskaeufler: If I say

```patch
  macro setup_initialize
    def initialize(
        {% for field in FIELDS %}
          @{{field[:name]}},
        {% end %}
      )
+      @id.foo
    end
  end
```

```
in macro 'setup_initialize' /Users/hanneskaeufler/Projects/forks/lucky/lucky_record/src/lucky_record/model.cr:62, line 18:

   1.     def initialize(
   2.
   3.           @id,
   4.
   5.           @created_at,
   6.
   7.           @updated_at,
   8.
   9.           @name,
  10.
  11.           @age,
  12.
  13.           @nickname,
  14.
  15.           @joined_at,
  16.
  17.       )
> 18.       @id.foo
  19.     end
  20.

undefined method 'foo' for Int32
```

paulcsmith: Oh I see. That's because the initializer accepts the raw column type. So in that case you'd do something like the spec here for CustomEmail: https://github.com/luckyframework/lucky_record/blob/ff9dde975d56dce436c77fe9bcef89a1c4ec4bad/spec/model_spec.cr#L84-L86

I *think* that will get you closer, but I'll try to take a closer look tonight :)

hanneskaeufler: It did get me closer 👍 , but I find that weird :) Why do I need to pass a String into the initializer if I have a UUID column?

paulcsmith: The DB.mapping passes those values to the initializer, so it takes the value the the DB.mapping uses. In this case, a String. 

It is a bit confusing, but in reality you should never use the initializer directly. It's only ever done in the specs, and even then, it is super rare

hanneskaeufler: Ahh, that explains it well, thanks

hanneskaeufler:
I think this might even work^^ But probably needs a bit playing around.

spec/form_spec.cr
```diff
@@ -363,6 +367,16 @@ describe "LuckyRecord::Form" do
...
+
+    context "with a uuid backed model" do
+      it "doesn't generate a new uuid" do
+        line_item = LineItemBox.create
+        LineItemForm.update(line_item, { "name" => "Another pair of shoes"}) do |form, record|
+          form.saved?.should be_true
+          record.id.should eq line_item.id
+        end
+      end
+    end
```
paulcsmith: Might be worth adding a test case for manually setting the id with a UUID as mentioned in the comment in form.cr: `UserForm.create!(id: UUID.new("some-hardcoded-uuid"))` should remain unchanged

spec/form_spec.cr
```diff
@@ -363,6 +367,16 @@ describe "LuckyRecord::Form" do
         end
       end
     end
+
```
paulcsmith: It might be worth adding a spec for creating a new line item with the LineItemForm to make sure the form is hooked up correctly

spec/type_extensions/uuid_spec.cr
```diff
@@ -0,0 +1,27 @@
+require "../../spec_helper"
+
+describe String do
```
paulcsmith: I think this should be "UUID column type"

spec/type_extensions/uuid_spec.cr
```diff
@@ -0,0 +1,27 @@
...
+    end
+  end
+
+  describe ".to_db" do
+    it "turns the uuid into a string" do
+      uuid = UUID.new("a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11")
+      UUID::Lucky.to_db(uuid).should eq "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11"
+    end
+  end
+end
```
paulcsmith: Very nice spec!

spec/model_spec.cr
```diff
@@ -1,4 +1,5 @@
 require "./spec_helper"
+require "uuid"
```
paulcsmith: I think this should be moved to `src/lucky_record.cr` so the requires are (mostly) in one place!

spec/form_spec.cr
hanneskaeufler: Agreed, I didn't think that manually setting ids was allowed. Test added ✅

hanneskaeufler:
Your comments are adressed, and for fun and giggles I migrated my blog to uuids (in dev with symlinked lucky_record) and ... omg that worked oO See https://github.com/hanneskaeufler/blog/pull/70 for needed changes.

paulcsmith:
Very cool!

I wonder how it works with foreign keys though. I just realized that may cause issues. For example, Comment that belongs to Post. Does querying and preloading still work I wonder. I'm not sure if LuckyRecord will know to use a UUID or if it will blow up :S

hanneskaeufler:
> I wonder how it works with foreign keys though.

Mhm, no idea. Let's see if I can get a test working.

paulcsmith:
Thank you! That's something I totally forgot about until just barely :P

src/lucky_record/associations.cr
```diff
@@ -172,7 +172,9 @@ module LuckyRecord::Associations
       {% nilable = false %}
     {% end %}
 
-    column {{ assoc_name.id }}_id : Int32{% if nilable %}?{% end %}
```
hanneskaeufler: So this hardcoded `Int32` was blowing up. I haven't worked with any lucky_record associations yet really, so I'm not aware of all the use cases. But I was able to fix this one by fetching the `id` column type from the "owning" model.

paulcsmith: This makes perfect sense 👍 

I think it may be more resilient and a bit faster to have a `PRIMARY_KEY_TYPE` constant on the model so you can shorten it to `model.resolve.constant(:PRIMARY_KEY_TYPE)` or something. What do you think?

hanneskaeufler: Checks out, thanks for the tip 👍

paulcsmith:
Another place to check would be for `has_many` and preloading in the `preloading_spec`. Maybe you're already on it, but I figured I'd comment just in case :)

hanneskaeufler:
Yeah I'm on it. Preloading `has_many` with `through:` remains. That's one for another day, tired of defining fixture classes for now lol.

src/lucky_record/associations.cr
```diff
@@ -72,7 +72,14 @@ module LuckyRecord::Associations
               end
               .preload_{{ through.id }}
               .distinct
-            {{ assoc_name }} = {} of Int32 => Array({{ model }})
+
+            {% owner_id_type = if model.id == "Product"
```
hanneskaeufler: So yeah @paulcsmith I'm currently stuck resolving this conditional, which should obviously be `    {% owner_id_type = model.resolve.constant(:PRIMARY_KEY_TYPE_CLASS) %}`, however that fails in a way I don't understand:

```
Error in line 1: while requiring "./spec/insert_spec.cr"

in spec/insert_spec.cr:1: while requiring "./spec_helper"

require "./spec_helper"
^

in spec/spec_helper.cr:3: while requiring "./support/**"

require "./support/**"
^

in spec/support/comment.cr:1: while requiring "./post"

require "./post"
^

in spec/support/post.cr:9: expanding macro

    ^~~~~~~~

in src/lucky_record/associations.cr:76: undefined constant Tag

            {% owner_id_type = model.resolve.constant(:PRIMARY_KEY_TYPE_CLASS) %}
                                     ^~~~~~~

in spec/support/post.cr:9: undefined constant Tag

```

Why does that suddenly not know the Tag class? I tried `require`-ing the models error by error but ultimately couldn't fix the undefined constants errors. Any ideas?

paulcsmith: That is definitely strange. I'll have to pull it down and check it out!

db/migrations/20180627220827_create_price_for_line_items.cr
```diff
@@ -0,0 +1,22 @@
+class CreatePriceForLineItems::V20180627220827 < LuckyMigrator::Migration::V1
+  def migrate
+    # Needs support in lucky_migrator
+    # create :prices, primary_key_type: :uuid do
+    #   add in_cents : Int32
+    #   add line_item_id : UUID
```
paulcsmith: For the comment here, could you create an issue on LuckyMigrator?

Also, I think it would be something like this:

```
add_belongs_to line_item : LineItem, foreign_key_type: UUID
```

So it can set up foreign keys and all that jazz as well

hanneskaeufler: Done, I just found that out as well. So I guess we can either leave this comment and merge, or wait until lucky_migrator has support ...

paulcsmith: I think removing the comment and adding support for it in LuckyMigrator later would be ok.

That way we can at least make it available for people now, and refine the experience later. We may even have time to add it in LuckyMigrator before release :)

hanneskaeufler: Sure, done

PR #257, opened 2018-06-29
Infers the table name, closes #231

### LINKED ISSUE #231: Infer table name automatically
```crystal
class User < BaseModel
  table do
    # fields
  end
end
```

Maybe even remove `table` macro:

```crystal
class User < BaseModel
  column name : String
  belongs_to etc.
end
```

I kind of like the `table` macro because it keeps all table specific stuff in one place.

hanneskaeufler:
Nothing fancy really, but allows you to omit the table name. Works for the simple 90% case e.g.

```crystal
class Post < LuckyRecord::Model
  table do
  end
end
```

which would result in `posts` as a table name.


hanneskaeufler:
@paulcsmith Switched to `.underscore` now, just discovered that. Also, travis doesn't run anymore now, was that enabled by accident?

paulcsmith:
@hanneskaeufler Yes Travis was enabled by accident when adding it to a different org.

I'll take a closer look later, but I think we may want to support more complex use-cases with by using `LuckyInflector`. Either that or maybe it is easier to make table names singular. I'm not sure if I like that idea though. What do you think?

hanneskaeufler:
I prefer pluralized table names for sure. No technical reason though, really. I was scared of introducing the inflector because I didn't know how to call that at compile time :D I can look into that though.

paulcsmith:
@hanneskaeufler It can be done, but it's a bit tricky. You have to use the `run` macro. You can see it in use in Lucky router:

* https://github.com/luckyframework/lucky/blob/500e4bc28786722700672c398b768b3937b204cd/src/lucky/routeable.cr#L128
* Which calls https://github.com/luckyframework/lucky/blob/master/src/run_macros/infer_route.cr

Important to note that the macros used by `run` are in a separate folder in `src/run_macros`. That's so they are not required by Lucky when it is loaded. They should only be called when using the `run` macro

paulcsmith:
And in this file you can see how we're generating singularized names for routes at compile time: https://github.com/luckyframework/lucky/blob/500e4bc28786722700672c398b768b3937b204cd/src/run_macros/infer_route.cr#L91

paulcsmith:
And some docs on the `run` macro https://crystal-lang.org/api/0.25.1/Crystal/Macros.html#run(filename,*args):MacroId-instance-method

shard.yml
```diff
@@ -18,6 +18,9 @@ dependencies:
     version: ~> 0.15
   habitat:
     github: luckyframework/habitat
+  lucky_inflector:
+    gitbug: luckyframework/lucky_inflector
```
hanneskaeufler: Tell me why did that `shard update` correctly locally? :D :D

hanneskaeufler:
Thanks paul, with your hints it was fairly easy to make this work.

src/lucky_record/model.cr
```diff
@@ -30,7 +30,8 @@ class LuckyRecord::Model
     id.to_s
   end
 
-  macro table(table_name, primary_key_type = LuckyRecord::PrimaryKeyType::Serial)
+  macro table(table_name = nil, primary_key_type = LuckyRecord::PrimaryKeyType::Serial)
+    {% table_name = run("../run_macros/infer_table_name.cr", @type.id) unless table_name %}
```
paulcsmith: I think the `unless` at the end of this kind of sneaks in and makes it hard to see. Could you move it to above? `{% unless table_name %}`?

hanneskaeufler:
Yupp, done

PR #259, opened 2018-07-02
has_many through with uuids

### LINKED ISSUE #258: UUID based has_many through
Due to my lack of undestanding a compile error in the specs, has_many through support is left out of #254.

hanneskaeufler:
Fixes #258


hanneskaeufler:
For minimal example to reproduce the tricky bit: https://play.crystal-lang.org/#/r/4ga1

```crystal
class Model
  macro inherited
    PRIMARY_ID_CLASS = Int32.id
  end
  
  macro relation(declaration)
    {% declaration.type.resolve.constant(:PRIMARY_ID_CLASS) %}
  end
end

class Tag < Model
  relation post : Post
end

class Post < Model
  relation tag : Tag
end
```

hanneskaeufler:
I _think_ this can be done. @paulcsmith mentioned the "finished" macro. I was able to get this to compile and output the expected:

```crystal
class Model
  macro inherited
    ASSOCIATIONS = [] of {type: Object}

    macro finished
      process_associations
    end
  end

  macro table(primary_key_type)
    PRIMARY_ID_CLASS = {{ primary_key_type }}
    {{yield}}
  end

  macro process_associations
    def print_assocs
      puts <<-TEXT
      {% for assoc in ASSOCIATIONS %}
        {{ assoc[:type] }} => {{ assoc[:type].resolve.constant(:PRIMARY_ID_CLASS) }}
      {% end %}
      TEXT
    end
  end

  macro relation(declaration)
    {% ASSOCIATIONS << {type: declaration.type} %}
  end
end

class Tag < Model
  table(Int32) do
    relation(post : Post)
  end
end

class Post < Model
  table(UUID) do
    relation(tag : Tag)
  end
end

Post.new.print_assocs
Tag.new.print_assocs
```
Output
```
  Tag => Int32


  Post => UUID
```

paulcsmith:
Very nice. Yeah I think something like that will make it work. I'll be on vacation next week, but I might have time to help with this. We'll see :)

hanneskaeufler:
So I beat this thing to death with a stick. Moving the offending `resolve` call into a `finished` macro did the trick. We _could_ move more of the associations setup into the finished step, but do we really need to? Dunno.

giphy-2

src/lucky_record/model.cr
```diff
@@ -9,6 +9,13 @@ class LuckyRecord::Model
   macro inherited
     FIELDS = [] of {name: Symbol, type: Object, nilable: Bool, autogenerated: Bool}
     ASSOCIATIONS = [] of {name: Symbol, foreign_key: Symbol, type: Object}
+    HAS_MANY = [] of {name: Object, foreign_key: Symbol, type: Object, through: Object}
```
hanneskaeufler: Definitely not loving storing the has_many associations one more time, but I think I need it. Do you see a cleaner way?

src/lucky_record/associations.cr
```diff
@@ -9,88 +9,20 @@ module LuckyRecord::Associations
...
-              .results.group_by(&.{{ foreign_key }})
-          {% end %}
-          records.each do |record|
-            record._preloaded_{{ assoc_name }} = {{ assoc_name }}[record.id]? || [] of {{ model }}
-          end
-        end
-        self
-      end
-    end
+    define_has_many_lazy_loading({{ assoc_name }}, {{ model }}, {{ foreign_key }}, {{ through }})
```
paulcsmith: I love these new macros. They read really nicely 👍

src/lucky_record/model.cr
paulcsmith: You could potentially store the association type in the ASSOCIATIONS constant. `{relationship_type: :has_many|:belongs_to|:has_one|:etc}`

hanneskaeufler: I did so, thanks. Tiny bit cleaner I think.

PR #263, opened 2018-07-13
Ignore self when validating uniqueness.

oneiros:
validates_uniqueness_of can be useful on update, in case the unique field's value is being updated. But if the value stays the same, the validation reported an error.

This is now fixed, as the current record is being ignored when searching for conflicting records in the database.

Of course, when supplying your own query, you are on your own :slightly_frowning_face: 

I dabbled with automatically injecting something in this case as well, but gave up after some time. I guess it is better to document this as a caveat than to tamper with the user supplied query.

### DIFF
--- a/src/lucky_record/validations.cr
+++ b/src/lucky_record/validations.cr
@@ -81,7 +81,7 @@ module LuckyRecord::Validations
     private def build_validation_query(column_name, value) : T::BaseQuery
       query = T::BaseQuery.new.where(column_name, value)
       record.try(&.id).try do |id|
-        query = query.where("#{T::TABLE_NAME}.id != ?", id)
+        query = query.id.not(id)
       end
       query
     end
--- a/src/lucky_record/validations.cr
+++ b/src/lucky_record/validations.cr
@@ -79,7 +79,13 @@ module LuckyRecord::Validations
   # if because there is no T (model class).
   macro included
     private def build_validation_query(column_name, value) : T::BaseQuery
-      T::BaseQuery.new.where(column_name, value)
+      query = T::BaseQuery.new.where(column_name, value)
+      record.try do |r|
+        r.id.try do |id|
+          query = query.where("#{T::TABLE_NAME}.id != ?", id)
+        end
+      end
+      query
     end
   end
 end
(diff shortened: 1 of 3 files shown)


oneiros:
I also opened an issue here (#264) in case you prefer a different solution.

src/lucky_record/validations.cr
```diff
@@ -79,7 +79,13 @@ module LuckyRecord::Validations
   # if because there is no T (model class).
   macro included
     private def build_validation_query(column_name, value) : T::BaseQuery
-      T::BaseQuery.new.where(column_name, value)
+      query = T::BaseQuery.new.where(column_name, value)
+      record.try do |r|
+        r.id.try do |id|
+          query = query.where("#{T::TABLE_NAME}.id != ?", id)
```
paulcsmith: I believe you can do `query.id.not(id)`

src/lucky_record/validations.cr
```diff
@@ -79,7 +79,13 @@ module LuckyRecord::Validations
   # if because there is no T (model class).
   macro included
     private def build_validation_query(column_name, value) : T::BaseQuery
-      T::BaseQuery.new.where(column_name, value)
+      query = T::BaseQuery.new.where(column_name, value)
+      record.try do |r|
+        r.id.try do |id|
```
paulcsmith: I think you can simplify this like so: 
```
record.try(&.id).try do |id|
  # do stuff with the query
end
```

src/lucky_record/validations.cr
paulcsmith: Yeah let's remove the `private` to get this working 👍. I think that should be fine.

Thanks for digging in and figuring that out!

oneiros:
> You can make it work with a custom query by doing the same query = query.id.not(id) here: https://github.com/luckyframework/lucky_record/pull/263/files#diff-14eb7e3c5870e9675ed5c0f7b03ee04dR58

I actually tried a variaton of this, but this will not compile either. There is a class `TestValidationUser` defined in `spec/validations_spec.cr` that includes `LuckyRecord::Validations` and calls this method. But it is not DB-backed and hence there is no `record` method to get an id from.

I guess this could be worked around to some degree, but I wondered if that is really necessary. I am not sure if one should mess with the custom query. As it stands now, this could be anything, even something relating to a totally different table in the database.

src/lucky_record/validations.cr
oneiros: OK. I just pushed another update.

Issue #238, opened 2018-06-12
Renaming LuckyRecord to something more generic


paulcsmith:
LuckyFlow will likely be renamed soon so that it is clearer that it can be used outside Lucky.

I think that might also apply to LuckyRecord. There is nothing in it that is tied specifically to Lucky so it makes sense to give it a more generic name so it's clear people can use it in their own frameworks and apps outside of Lucky.

What are your thoughts on this?

## Possible names

* Atomic
* Bokeh
* Turbine
* Filament
* Anvil

Would love to hear your thoughts!

hfjallemark:
As mentioned in the Gitter channel; +1 for abstract names 😃 Would be cool if the name could somehow tie into the use case for the library (not that I have any good suggestions).

FWIW from your list I like Atomic and Turbine 👍

neurodynamic:
I’m very pro taking a leaf from the Elm community book and naming it after what it does. Or coming as close as possible. HelpfulORM or HelpfulRecord or something maybe?

robacarp:
A lot of these names will struggle with search-ability. I like Bokeh, for example, but what are people going to type into a search engine? I enjoy the culture of abstract names but it often hinders usability and search-ability. 

I like the idea of using a pair of words, like @neurodynamic suggested. I did have a bad experience removing an ORM suffix once though...

paulcsmith:
Those are some good points. I think there are some good counterpoints too. Usually a person would pair the library name with something else that helps narrow things down, for example: "Bokeh virtual form" or "Bokeh distinct query". For exampleThe same worry could be said for Rails, node, etc., but Google and most search engines learn about these word pairing and tend to get things right once it has had a few months to learn search behavior and categorize things as "programming"

The downside of more concrete names is that you run out of names quickly. There are a number of ORMs right now and I don't think there are enough adjectives to use for everyone's libraries `HelpfulRecord`, `FastRecord`, etc.

I think abstract names are nice because:

1) They're fun :D But this is purely subjective. I know lots of people think it is annoying
2) It allows for multiple Crystal ORMs

So maybe a mix of abstract and concrete names to get the best of both worlds? "#{Abstract name} ORM". Example: "Bokeh ORM", "Plucky ORM", etc.

Thoughts?

jayroh:
I'm quite liking "Clover".  It's distinct on it's own but still carries the "Lucky" DNA it was born from.

"Clover ORM" / "CloverRecord" / something along those lines.

[...]

PR #269, opened 2018-07-17
Pass Int32 Values to DB as integer

mikeeus:
I used `hub` to get the old pull request and rebase it (awesome tool btw). There are a couple issues though:

1. The changes made for UUID types breaks the `crystal tasks.cr db.reset` task because for whatever reason running `shards` doesn't install the latest version of `LuckyMigrator`.
```
in db/migrations/20180628193054_create_product.cr:3: no argument named 'primary_key_type'

    create :products, primary_key_type: :uuid
```

2. I had to make the `where` clause accept `Nil` to allow `validates_uniqueness_of` to work since it can potentially send a `nil` value.
```
private def validate_uniqueness_of(
    field : LuckyRecord::Field,
    message : String = "is already taken"
  )
    field.value.try do |value|
      if build_validation_query(field.name, field.value).first? # field.value might be nil
        field.add_error message
      end
    end
  end
```
This wasn't a problem before because nil would be converted into an empty string. Do you think allowing `Nil` is okay or should I deal with it differently?

### DIFF
--- a/spec/criteria_spec.cr
+++ b/spec/criteria_spec.cr
@@ -11,54 +11,54 @@ end
 describe LuckyRecord::Criteria do
   describe "is" do
     it "uses =" do
-      age.is(30).to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE users.age = $1", "30"]
+      age.is(30).to_sql.should eq ["SELECT #{QueryMe::COLUMNS} FROM users WHERE users.age = $1", 30]
     end
   end
 
   describe "is_not" do
     it "uses !=" do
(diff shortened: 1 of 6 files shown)


edwardloveall:
Hey @mikeeus! We've moved and [renamed] LuckyRecord to [Avram]. We decided to move it instead of using GitHub's rename option so that current code that relies on lucky_record would continue to work and a project must opt-in to the new shard. However, a big bummer is we can't move PRs.

If you are still interested in this change, would you be able to rebase onto Avram? If not don't worry about it. We can eventually rebase it for you. Thanks 😄!

PR #272, opened 2018-07-27
Fix time parsing

edwardloveall:
Crystal recently updated how it parses dates, but they also added a 
nice 
convenience method that we can use to parse strings since we expect 
them 
to be UTC.

### DIFF
--- a/spec/type_extensions/time_spec.cr
+++ b/spec/type_extensions/time_spec.cr
@@ -1,0 +1,29 @@
+require "../../spec_helper"
+
+describe "Time column type" do
+  describe ".parse" do
+    it "casts an ISO8601 String successfully" do
+      time = Time.new.to_s("%FT%X%z")
+
+      result = Time::Lucky.parse(time)
+
+      result.should be_a(LuckyRecord::Type::SuccessfulCast(Time))
+    end
+
+    it "casts a Time successfully" do
+      time = Time.new
+
+      result = Time::Lucky.parse(time)
+
+      result.value.should eq(time)
+    end
+
+    it "can't cast an invalid value" do
+      time = Time.new
+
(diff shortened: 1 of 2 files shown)


edwardloveall:
We do have this: https://github.com/luckyframework/lucky_record/blob/master/spec/form_spec.cr#L146-L152

Do you also want a more unit-y test?

paulcsmith:
Yeah I think a more unit-y test would be good since that spec didn't seem to fail. So if you revert this change, whatever spec is added should ideally fail. That way we know for sure we won't have a future regression

edwardloveall:
Added!

spec/type_extensions/time_spec.cr
```diff
@@ -0,0 +1,29 @@
+require "../../spec_helper"
+
+describe "UUID column type" do
```
paulcsmith: Time type :)

spec/type_extensions/time_spec.cr
```diff
@@ -0,0 +1,29 @@
...
+  describe ".parse" do
+    it "casts a String successfully" do
+      time = Time.new
```
paulcsmith: I'd do the `to_s("etc.")` here so it is in the setup phase. What do you think about that?

spec/type_extensions/time_spec.cr
```diff
@@ -0,0 +1,29 @@
+require "../../spec_helper"
+
+describe "UUID column type" do
+  describe ".parse" do
+    it "casts a String successfully" do
```
paulcsmith: Maybe it could say what kind of string? Like `casts an ISO8601 string successfully`?

edwardloveall:
Updated for your comments

PR #275, opened 2018-08-02
Add jsonb support

tzar:
I have added some rudimentary (parse/load/save) support for jsonb as a column type, but it will need https://github.com/luckyframework/lucky_migrator/pull/122 for the specs to succeed as it needs to `add` a `JSON::Any` column.

I haven't worked much with either lucky or crystal in the past, so let me know if there's anything I can improve here


tzar:
Added the jsonb column to test against using `execute` instead so specs can run

tzar:
I need to rework to not use `JSON::Type` as it's gone in crystal v0.25

spec/json_column_spec.cr
```diff
@@ -0,0 +1,62 @@
...
+
+class BlobForm < Blob::BaseForm
+  fillable :doc
+end
+
+describe "JSON Columns" do
+  it "should work in boxes" do
+    BlobBox.create
+    blob = BlobQuery.new.first
+    blob.doc.should eq JSON::Any.new({"foo" => JSON::Any.new("bar")})
```
mikeeus: Would `({"foo" => "bar"}).to_json` work here as well? If so I think it would be more readable since calling `JSON::Any.new` is verbose. Also you can convert NamedTuples too like `({ foo: "bar" }).to_json`

src/lucky_record/charms/json_extensions.cr
```diff
@@ -3,6 +3,8 @@ struct JSON::Any
     alias ColumnType = JSON::Any
     include LuckyRecord::Type
 
+    alias JSONType = Array(JSONType) | Bool | Float64 | Hash(String, JSONType) | Int64 | String | Nil | Int32 | Float32
```
mikeeus: I *think* you can add `NamedTuple(Symbol, JSONType)` if you use `to_json` to do the conversions

mikeeus:
This is looking really good! I added a couple comments about using `to_json` and parsing `NamedTuples`. I think @paulcsmith would have some comments to go over then it should be good to go.

tzar:
I will try find time to merge https://github.com/luckyframework/lucky_migrator/pull/122 into this PR next week :)

mikeeus:
@tzar Awesome work on this! For the travis config, have you tried adding:
```
addons:
  postgresql: "9.4"
```
to `.travis.yml`?

tzar:
Thanks! I assumed I wouldn't be able to mess with travis from within the PR, but I should have just tried it ☺

edwardloveall:
Hey @tzar! We've moved and [renamed] LuckyRecord to [Avram]. We decided to move it instead of using GitHub's rename option so that current code that relies on lucky_record would continue to work and a project must opt-in to the new shard. However, a big bummer is we can't move PRs.

tzar:
Done! Closing this one in favour of https://github.com/luckyframework/avram/pull/48

PR #283, opened 2018-08-28
Upgrade habitat to 0.4.0

paulcsmith:


paulcsmith:
@fernandes Great question. Previously it was using Circle CI and I believe it already had a db created. It looks like on Travis it needs it to be explicit, but I'm not sure. Still trying to get it set up :D

fernandes:
@paulcsmith hahahah that's right, travis comes with no DB

What I do usually (like testing a rails application), is similar to this:

```yaml
before_script:
  - psql -c 'create database database_name;' -U travis
```

and works, but `crystal tasks.cr -- db.create` is more elegant in this case heheh

fernandes:
@paulcsmith what's happening on travis build:

on [line](https://travis-ci.com/luckyframework/lucky_record/jobs/142542484#L585) shows your command `db.create` is creating a `lucky_record_dev` because running as dev.

And [spec_helper](https://github.com/luckyframework/lucky_record/blob/a74caf9217d0bb9d5b6940c171a2751fcf97899d/spec/spec_helper.cr#L11) tries to connect to `lucky_record_test`, and that's why `crystal spec` is failing.

Adding right before crystal spec would do the job:

```
DATABASE_URL='psql://127.0.0.1:5432/lucky_record_test' crystal tasks.cr -- db.create
```

DATABASE_URL is just for [MigratorRunner](https://github.com/luckyframework/lucky_record/blob/a74caf9217d0bb9d5b6940c171a2751fcf97899d/src/lucky_record/migrator/runner.cr#L16) choose the right `db_name` to created

PR #289, opened 2018-10-04
Fix comparisons to `nil` in queries & make such queries explicit

### LINKED ISSUE #123: Disallow `nil` when querying. Require using `is_nil` instead
If you want to allow a value *or* nil maybe have a method like `is?(something_that_might_be_nil)`. There may be some other name that is better. Will have to think about it a bit

hibachrach:
Fixes #123

Not totally happy with `is?`/`is_not?` as method names as it overloads a third additional meaning for `?`-suffixed methods (the first two being (1) methods which can return `nil` and (2) methods which always return booleans) so I'm open to ideas there.

### DIFF
--- a/src/lucky_record/where.cr
+++ b/src/lucky_record/where.cr
@@ -25,7 +25,6 @@ module LuckyRecord::Where
     end
   end
 
-
   class Null < NullSqlClause
     def operator
       "IS"
--- a/src/lucky_record/where.cr
+++ b/src/lucky_record/where.cr
@@ -1,19 +1,53 @@
 module LuckyRecord::Where
   abstract class SqlClause
-    getter :column, :value
+    getter :column
+
+    def initialize(@column : Symbol | String)
+    end
+
+    abstract def operator : String
+    abstract def negated : SqlClause
+
+    def prepare
+      "#{column} #{operator}"
+    end
+  end
+
+  abstract class ComparativeSqlClause < SqlClause
+    getter :value
 
     def initialize(@column : Symbol | String, @value : String | Array(String) | Array(Int32))
     end
 
     abstract def operator : String
-    abstract def negated : SqlClause
+    abstract def negated : ComparativeSqlClause
 
     def prepare(prepared_statement_placeholder : String)
       "#{column} #{operator} #{prepared_statement_placeholder}"
     end
   end
 
-  class Equal < SqlClause
+  class Null < SqlClause
+    def operator
+      "IS NULL"
+    end
+
+    def negated : NotNull
+      NotNull.new(@column)
+    end
+  end
+
+  class NotNull < SqlClause
+    def operator
+      "IS NOT NULL"
+    end
+
+    def negated : Null
+      Null.new(@column)
+    end
+  end
+
+  class Equal < ComparativeSqlClause
     def operator
       "="
     end
@@ -23,7 +57,7 @@ module LuckyRecord::Where
     end
   end
 
-  class NotEqual < SqlClause
+  class NotEqual < ComparativeSqlClause
     def operator
       "!="
     end
@@ -33,7 +67,7 @@ module LuckyRecord::Where
     end
   end
 
-  class GreaterThan < SqlClause
+  class GreaterThan < ComparativeSqlClause
     def operator
       ">"
     end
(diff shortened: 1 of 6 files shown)


src/lucky_record/where.cr
```diff
@@ -1,19 +1,53 @@
...
 
     abstract def operator : String
     abstract def negated : SqlClause
 
+    def prepare
+      "#{column} #{operator}"
+    end
+  end
+
+  abstract class ComparativeSqlClause < SqlClause
```
mikeeus: I think I understand why you introduced this class. Inheriting from `SqlClause` and overriding the constructor to only accept the `@column` will throw an error unless you make `@value` nilable, which would break the class.

Your solution does work well to separate the two concepts, but I think clauses that reference `NULL` are a special case and should inherit from `SqlClause` instead of the other way around.

You can achieve this without changing `SqlClause` by introducing a `NullSqlClause` class and removing `@value` from the constructor by setting it to `"NULL":

```crystal
abstract class SqlClause
  getter :column, :value

  def initialize(@column : Symbol | String, @value : String | Array(String) | Array(Int32))
  end

  abstract def operator : String
  abstract def negated : SqlClause

  def prepare(prepared_statement_placeholder : String)
    "#{column} #{operator} #{prepared_statement_placeholder}"
  end
end

abstract class NullSqlClause < SqlClause
  @value = "NULL"

  def initialize(@column : Symbol | String)
  end
end
```
The `Null` and `NotNull` classes can then use the default `prepare` method to generate their sql so you can set their operators to "IS" and "IS NOT":

```crystal
class Null < NullSqlClause
  def operator
    "IS"
  end
  ...
end
```
I believe this also means you can use the normal `prepared_statement_values` and `joined_wheres_queries` methods.

mikeeus (commented):
Thanks a lot for working on this! All the specs are set up and its looking good except for the comment I made about the `LuckyRecord::Where::ComparativeSqlClause` class. I would love to hear your thoughts on it :)

src/lucky_record/where.cr
HarrisonB: Hmm yeah I like this idea better. It's likely to be the only clause of its sort.

HarrisonB: One thing: I'm not sure how you can get `NullSqlClause` to use the normal `prepared_statement_values` as that would result with `users.nickname IS NULL $1`.

mikeeus: If you change the operator to "IS" it should come out like `users.nickname IS $1` and work  as expected, right?

HarrisonB: Ah yes. Sorry!

HarrisonB: Sorry to keep asking questions about this, but won't passing in `NULL` as a dollar-sign param. result in it getting escaped? And just treated like the string `"NULL"`? And we can't override the ivar in the child class to set it to a `nil` literal.

mikeeus: It's no problem, I missed the details and oversimplified it. I think you can do with another prepare method then :)
```
def prepare
   "#{column} #{operator} #{@value}"
end
```

paulcsmith:
@HarrisonB Thank you for working on this! I still need to look at the code, but I am happy you proposed some new syntaxes. As you discussed here and in https://gitter.im/luckyframework/Lobby?at=5c32c11c6370df0b919feb27 I think `?` is a bit ambiguous. I do like changing `is` -> `eq` a lot.

I think we should go with:

```
is -> eq
is_not -> not_eq
is? -> nilable_eq
is_not? -> not_nilable_eq
```

And actually I'm wondering if we even need `is_not` since we have `.not.eq` and `.not.nilable_eq`. 

Thoughts on removing the `*_not` version? Any other thoughts on this?

edwardloveall:
Hey @HarrisonB! We've moved and [renamed] LuckyRecord to [Avram] (thanks for the name idea!). We decided to move it instead of using GitHub's rename option so that current code that relies on lucky_record would continue to work and a project must opt-in to the new shard. However, a big bummer is we can't move PRs.

hibachrach:
Certainly!

I’m a bit surprised GitHub doesn’t just have a simple rename functionality for repositories.

Hopefully it’s just a one-time thing haha

paulcsmith:
Yeah there is a rename function but I’ve heard that it sometimes causes weird issues with got remotes and messed up links :S

And yes this won’t happen again!

PR #291, opened 2018-10-06
Add `Queryable#distinct_on`

### LINKED ISSUE #255: Implement distinct_on
Lucky record currently handles distinct but not distinct_on. We should be able to do something to reflect this:
```
select distinct on (name)
    name, col1, col2
from names
order by name, col1
```

HarrisonB:
Closes #255.

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -23,10 +23,22 @@ describe LuckyRecord::Query do
   end
 
   it "can select distinct on a specific column" do
-    query = UserQuery.new.distinct_on(&.name).query
+    UserBox.new.name("Purcell").age(22).create
+    UserBox.new.name("Purcell").age(84).create
+    UserBox.new.name("Griffiths").age(55).create
+    UserBox.new.name("Griffiths").age(75).create
+    queryable = UserQuery.new.distinct_on(&.name).order_by(:name, :asc).order_by(:age, :asc)
+    query = queryable.query
 
-    query.statement.should eq "SELECT DISTINCT ON (users.name) #{User::COLUMNS} FROM users"
+    query.statement.should eq "SELECT DISTINCT ON (users.name) #{User::COLUMNS} FROM users ORDER BY name, age ASC"
     query.args.should eq [] of String
+    results = queryable.results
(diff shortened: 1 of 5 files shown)


src/lucky_record/queryable.cr
```diff
@@ -47,6 +47,12 @@ module LuckyRecord::Queryable(T)
     self
   end
 
+  def distinct_on(&block)
+    criteria = yield self
+    criteria.distinct_on
```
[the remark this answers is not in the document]
HarrisonB: Haha it was @paulcsmith’s idea 😛 you can thank him!

HarrisonB:
Thanks!

src/lucky_record/query_builder.cr
```diff
@@ -71,6 +72,12 @@ class LuckyRecord::QueryBuilder
     self
   end
 
+  def distinct_on(column)
+    @distinct = true
```
paulcsmith: Rather than setting @distinct to true, you could create a new private method called distinct? That returns true if @distinct is set or if @distinct_on is set. That way the logic for determining whether to add a distinct query is in one place

src/lucky_record/query_builder.cr
```diff
@@ -71,6 +72,12 @@ class LuckyRecord::QueryBuilder
     self
   end
 
+  def distinct_on(column)
```
paulcsmith: Can you add a type restriction for string | symbol so that someone doesn’t accidentally pass nil to it?

spec/query_spec.cr
```diff
@@ -22,6 +22,13 @@ describe LuckyRecord::Query do
     query.args.should eq [] of String
   end
 
+  it "can select distinct on a specific column" do
+    query = UserQuery.new.distinct_on(&.name).query
+
+    query.statement.should eq "SELECT DISTINCT ON (users.name) #{User::COLUMNS} FROM users"
+    query.args.should eq [] of String
```
paulcsmith: Could you run the query by calling “results” on the query to ensure the SQL is correct and it works with Postgres?

PR #293, opened 2018-10-07
added script to run tests and related commands

ajwann:
I'm not sure if it's my setup locally or not, but all the specs fail for me.

Edit: It's definitely my local setup.

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -1,0 +1,8 @@
+#! /bin/bash
+
+printf "\nrunning specs with 'crystal spec'\n\n"
+crystal spec
+printf "\nbuilding shards with 'shards build'\n\n"
+shards build
+printf "\nformatting code with 'crystal tool format src spec'\n\n"
+crystal tool format src spec
--- a/README.md
+++ b/README.md
@@ -22,9 +22,11 @@ require "lucky_record"
 
 1. Fork it ( https://github.com/luckyframework/lucky_record/fork )
 2. Create your feature branch (git checkout -b my-new-feature)
-3. Commit your changes (git commit -am 'Add some feature')
-4. Push to the branch (git push origin my-new-feature)
-5. Create a new Pull Request
+3. Make your changes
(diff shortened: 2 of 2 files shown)


bin/test
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+
+printf "\nrunning specs with 'crystal spec'\n\n"
+crystal spec
+printf "\nbuilding shards with 'shards build'\n\n"
+shards build
+printf "\nformatting code with 'crystal tool format src spec'\n\n"
+crystal tool format src spec
```
mikeeus: I think this is a good change! Just two thoughts:

- I think crystal tool format should be run on all the code instead of just the specs since the CI specs will fail if there are any changes.
- And do you think we should have `crystal spec` run at the end in case `crystal tool format` makes changes that break the specs?

ajwann: I'll tweak the script to it runs ```crystal tool format``` on the entire project, and I can see the value in running ```crystal spec``` as the final command. I'm thinking of a scenario when that would happen though. 

It would probably be a case where a new contributor makes a change and writes some new specs with formatting issues, runs ```./bin/test``` and then gets failures on the final ```crystal spec``` command. This sounds like a valid scenario and I think we would want the specs to fail in that case.

ajwann: Actually I'm thinking that we wouldn't want to run the formatter on the lib directory. Thoughts?

ajwann:
Also, this is for issue #593 in the main Lucky repo.

bin/test
mikeeus: You're right we don't want to run the formatter on the lib directory, but the `crystal tool format` command excludes the lib directory by default. You can see [the code here](https://github.com/crystal-lang/crystal/blob/e11ee26c7cfc5c1e1982e84d318af4d063a8de72/src/compiler/crystal/command/format.cr#L18)

PR #302, opened 2018-11-13
Allow adding UUID columns (that are not associations)

hanneskaeufler:
Adding a simple UUID column failed with "undefined method add_column"
because the UUID type was not recognized for default values.

A static default for UUIDs is kind of ... unintuitive ... anyway, but this gets me going for now.
What I think should be the next step is to maybe allow passing a block or something to `fill_existing_with: ...` in order to get a different value for each entry in the db. But I see that as a separate task to tackle?

(Storing a UUID in postgres can be done in UUID, but _generating_ one requires an extension, so I opted out of something like `DEFAULT generate_uuid_v1()`).

### DIFF
--- a/spec/migrator/alter_table_statement_spec.cr
+++ b/spec/migrator/alter_table_statement_spec.cr
@@ -13,6 +13,7 @@ describe LuckyRecord::Migrator::AlterTableStatement do
       add joined_at : Time, default: :now
       add updated_at : Time, fill_existing_with: :now
       add future_time : Time, default: Time.new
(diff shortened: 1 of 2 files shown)


edwardloveall:
Hey @hanneskaeufler! We've moved and [renamed] LuckyRecord to [Avram]. We decided to move it instead of using GitHub's rename option so that current code that relies on lucky_record would continue to work and a project must opt-in to the new shard. However, a big bummer is we can't move PRs.

hanneskaeufler:
Moved this into avram: https://github.com/luckyframework/avram/pull/47

PR #310, opened 2019-01-20
Add Queryable#empty?

edwardloveall:
I find that I'm looking for this functionality when I want to show
something on a page based on the presence of a record. For example, an
empty-state UI that shows "No posts yet" as opposed to listing out all
posts. This could check a PostQuery.new.empty? to determine whether to
loop through posts or show a sad face.

I also thought about checking for `first?` but that returns a record if
it's present. That could be worked around with `!...nil?` but at that
point `size.zero?` seemed like a pretty solid choice. Also, maybe it's
faster?

### DIFF
--- a/spec/query_spec.cr
+++ b/spec/query_spec.cr
@@ -289,6 +289,12 @@ describe LuckyRecord::Query do
     end
   end
 
+  describe "#empty?" do
+    query = UserQuery.new
+
+    query.empty?.should be_true
+  end
+
   describe "#none" do
     it "returns 0 records" do
       UserBox.create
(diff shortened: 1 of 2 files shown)


edwardloveall:
Hey me! We've moved and [renamed] LuckyRecord to [Avram]. We decided to move it instead of using GitHub's rename option so that current code that relies on lucky_record would continue to work and a project must opt-in to the new shard. However, a big bummer is we can't move PRs.

P.S. This is pretty silly of me to leave a comment for myself, but hey, it only seems fair.

PR #312, opened 2019-01-22
Corrects macro that generates create method of forms

### LINKED ISSUE #311: Forms create methods are generated incorrectly when using virtual attributes together with fillable attributes.
Hello, i just found this issues on the create form

1. `VIRTUAL_FIELDS` are not properly passed to the form inside the method.
They are defined here as arguments:
https://github.com/luckyframework/lucky_record/blob/6dfa3e2f801bf810cd4998ec36c9a53f51ab30c2/src/lucky_record/needy_initializer_and_save_methods.cr#L66-L79

But they are not assigned to the form.

2. FIELDS constant used on create method is different on BaseForm and ModelForm:
`Model::BaseForm`: FIELDS have all fields from model.
`ModelForm < Model::BaseForm.`: FIELDS is not defined.

I found this by testing with macros.

3. FIELDS constant is used to generate create methods for the form, so it generates wrong code:
`Model::BaseForm`: generate a create method with all fields.
`ModelForm < Model::BaseForm.`: generate a create method with none field because its not defined.

My solution 
* Assign properly the `VIRTUAL_FIELDS` to the form.
* Set the `FIELDS` constant on inherited forms.
* Create a new constant called `FILLABLE_FIELDS` for generating create methods. Add each field to the constant on the `fillable` macro.

I have a question too:
* Do `NEEDS_ON_CREATE`, `NEEDS_ON_UPDATE`, `NEEDS_ON_INITIALIZE`  from `needs` macro actually needed? I think they complicate the code unnecessarily. Also if somebody define a field with  `fillable` and also with `needs`, it will generate a wrong create method even with my solution.

I will post here the PR in a couple of hours.

Xosmond:
Fixes #311

### DIFF
--- a/src/lucky_record/base_form_template.cr
+++ b/src/lucky_record/base_form_template.cr
@@ -3,6 +3,8 @@ class LuckyRecord::BaseFormTemplate
     class BaseForm < LuckyRecord::Form({{ model_type }})
       macro inherited
         FOREIGN_KEY = "{{ model_type.stringify.underscore.id }}_id"
+        FIELDS          = {{ fields }}
+        FILLABLE_FIELDS = [] of String
       end
 
       def table_name
--- a/src/lucky_record/form.cr
+++ b/src/lucky_record/form.cr
@@ -69,7 +69,8 @@ abstract class LuckyRecord::Form(T)
   end
 
   macro add_fields(primary_key_type, fields)
-    FIELDS = {{ fields }}
+    FIELDS          = {{ fields }}
+    FILLABLE_FIELDS = {{ fields.map { |field| field[:name].id } }}
 
     private def extract_changes_from_params
       fillable_params.each do |key, value|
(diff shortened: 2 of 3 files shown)


edwardloveall:
Hey @Xosmond! We've moved and [renamed] LuckyRecord to [Avram]. We decided to move it instead of using GitHub's rename option so that current code that relies on lucky_record would continue to work and a project must opt-in to the new shard. However, a big bummer is we can't move PRs.

Xosmond:
@edwardloveall Hey sure, will do. But actually I wanted to know your opinion and also @paulcsmith about two ways for creating records.

This PR would fix the bug but also allow to create records like this (having `username` is a regular field and `password` and`password_confirmation` are virtual fields):
```crystal
RegistrationForm.create! username: "hello", password: "my password", password_confirmation: "my password"
```

But only fixing virtual parameters bug, would need us to write it like this:

```crystal
RegistrationForm.create! params: { "username" => "hello" }, password: "my password", password_confirmation: "my password"
```

So which do you think would be better?

paulcsmith:
I love that first version. Looks more type safe and I think it’s more intuitive! 👍

Xosmond:
@paulcsmith Ok i will move this and create the tests but I see NEEDS_ON_INITIALIZE, NEEDS_ON_CREATE and NEEDS_ON_UPDATE are not actually used, do you have a reason for these constants? I ask this because they can generate possible bugs.

